PR #11, opened 2019-09-23
Clarified bootnode content

MadelineMurray:
Clarified bootnode content


docs/HowTo/Deploy/Bootnodes.md
```diff
@@ -25,7 +25,7 @@ We recommend bootnode configuration is stored under source control.
 To allow for failure, specify all bootnodes on the command line (even to the bootnodes themselves). 
 
 !!! example 
-    If your network has 2 bootnodes, pass the following parameter to all nodes including the bootnodes. 
+    If your network has two bootnodes, pass the following parameter to all nodes, including the bootnodes. 
    
     `--bootnodes=enode://<publicKeyBootnode1>@10.0.0.100:30303, <publicKeyBootnode2>@10.0.1.101:30303`
```
joshuafernandes: replace 10.0.0.100 with <ip_of_bootnode> & same for bootnode ip2

docs/HowTo/Find-and-Connect/Bootnodes.md
```diff
@@ -1,43 +1,43 @@
...
-    * No arguments for the [`--bootnodes` option](../../Reference/CLI/CLI-Syntax.md#bootnodes) because this is your bootnode.
+2. Start the node specifying the genesis file and data directory. 
     
     !!! example
         ```
-        besu --genesis-file=privateNetworkGenesis.json --data-path=nodeDataPath --bootnodes
-         ```
-     
+        besu --genesis-file=privateNetworkGenesis.json --data-path=nodeDataPath
+        ```
```
joshuafernandes: Optionally add, if using more than 1 bootnode, specify all bootnodes here with `--bootnodes=enode://<publicKeyBootnode1>@<ip1>:30303,enode://<publicKeyBootnode2>@<ip2>:30303,...`

docs/Tutorials/Private-Network/Create-Private-Clique-Network.md
```diff
@@ -124,16 +124,15 @@ In `extraData`, replace `<Node 1 Address>` with the [address for Node-1](#2-get-
...
 ```
 
 ```bash tab="Windows"
-besu --data-path=data --genesis-file=..\cliqueGenesis.json --bootnodes --network-id 123 --rpc-http-enabled --rpc-http-api=ETH,NET,CLIQUE --host-whitelist="*" --rpc-http-cors-origins="all"    
+besu --data-path=data --genesis-file=..\cliqueGenesis.json --network-id 123 --rpc-http-enabled --rpc-http-api=ETH,NET,CLIQUE --host-whitelist="*" --rpc-http-cors-origins="all"    
 ```
 
 The command line specifies: 
 
-* No arguments for the [`--bootnodes`](../../Reference/CLI/CLI-Syntax.md#bootnodes) option because this is your bootnode
```
joshuafernandes: Id leave this in saying its using the default config value of [ ] implying it is a bootnode?

docs/HowTo/Deploy/Bootnodes.md
EdJoJob: the space after the comma should also be removed

docs/HowTo/Find-and-Connect/Bootnodes.md
NicolasMassart: But excluding the current bootnode? I don't think it's useful to list its own enode address in the list.

hmijail: AFAIU, it is not useful but it is harmless - so having the same list of bootnodes across all bootnodes might ease deployment. 
If this is right, then I think that it is important to state that listing the node's own address is useless but harmless.

docs/Tutorials/Private-Network/Create-Private-Clique-Network.md
hmijail: But why do that? Does it make a difference?

MadelineMurray: @joshuafernandes - I removed this because specifying `--bootnodes` and leaving the bootnodes option completely have the same effect (empty list of bootnodes) and users were getting confused thinking a bootnode was different to a regular node (this is the case for Geth).  In cases with more than one bootnode, a user then thought we didn't have any bootnodes because there wasn't a node with only `--bootnodes` 

Does that make sense?  Let me know if you want to discuss.

joshuafernandes: All good :) , whatever makes it easier for people to follow and use

docs/HowTo/Find-and-Connect/Bootnodes.md
MadelineMurray: Per @EdJoJob and @joshuafernandes we recommend having the same list of bootnodes for all nodes including the bootnodes.

hmijail: Recommend - but, why? What is the purpose of the recommendation? What would happen if not?

hmijail: A recommendation without explanations end up in superstition that might outlive its usefulness

MadelineMurray: Because it's easier to have the same bootnodes list for all nodes that have to remove each node from it's own list.  I'll add a note stating a node ignores itself in the list.

hmijail: So that's what I meant with "useless but harmless". Thank you!

PR #18, opened 2019-09-25
Added genesis file content 

MadelineMurray:
Added genesis file content


docs/Reference/Config-Items.md
```diff
@@ -16,6 +20,26 @@ Network configuration items are specified in the Hyperledger Besu genesis file.
 | `evmStackSize`      | Maximum stack size. Specify to increase the maximum stack size in private networks with very complex smart contracts. Default is `1024`. |
 
 
+## Genesis Block Parameters  
+ 
+The purpose of some genesis block parameters parameters varies depending on the consensus protocol
```
NicolasMassart: duplicated "parameters" word

docs/Reference/Config-Items.md
```diff
@@ -16,6 +20,26 @@ Network configuration items are specified in the Hyperledger Besu genesis file.
...
+
+* `difficulty`
+* `extraData`
+* `mixHash`
+ 
+Genesis block parameters with the same purpose across all consensus protocols are described below. 
+
+| Item                | Description                                                                                                                              |
+|---------------------|-:----------------------------------------------------------------------------------------------------------------------------------------|
+| `coinbase`          | Address to which mining rewards are paid. Can be any value in the genesis block .
```
NicolasMassart: Perhaps we could advertise the usually used `0x0000000000000000000000000000000000000000` value so users don't wonder which one to use as they can put whatever they want.

docs/Reference/Config-Items.md
```diff
@@ -16,6 +20,26 @@ Network configuration items are specified in the Hyperledger Besu genesis file.
...
+* `extraData`
+* `mixHash`
+ 
+Genesis block parameters with the same purpose across all consensus protocols are described below. 
+
+| Item                | Description                                                                                                                              |
+|---------------------|-:----------------------------------------------------------------------------------------------------------------------------------------|
+| `coinbase`          | Address to which mining rewards are paid. Can be any value in the genesis block . 
+| `gasLimit`          | Block gas limit. Total gas limit for all transactions in a block.                                                                                       |
+| `nonce`             | Used in block computation. Can be any value in the genesis block.
```
NicolasMassart: Perhaps we could advertise the usually used `0x0` value so users don't wonder which one to use as they can put whatever they want.

docs/Reference/Config-Items.md
```diff
@@ -16,6 +20,26 @@ Network configuration items are specified in the Hyperledger Besu genesis file.
 | `evmStackSize`      | Maximum stack size. Specify to increase the maximum stack size in private networks with very complex smart contracts. Default is `1024`. |
 
 
+## Genesis Block Parameters
```
lucassaldanha: Is `Block Parameters` the real name or is this something that we came up with?

Also, I think it would be useful to add an example of the meaning of these parameters in one of the consensus algorithms used (are we explaining them somewhere else?). E.g:
```
When using PoW, the difficulty represents how hard it is for a miner to create a block. The lower the value, the faster the node will generate blocks.
```

docs/Reference/Config-Items.md
MadelineMurray: Fixed

docs/Reference/Config-Items.md
MadelineMurray: Done

docs/Reference/Config-Items.md
MadelineMurray: Done

docs/Reference/Config-Items.md
MadelineMurray: The protocol names are linked and explanations are included there

MadelineMurray: Block Parameters used to differentiate those items from the config items

PR #39, opened 2019-10-01
Disable edit icon because of DCO issues

NicolasMassart:
## PR description
Disable edit icon because of DCO issues (see ticket)

Disable only instead of remove to be able to keep the same template with all our
sites and only comment config meta value.

## Fixed Issue(s)

fix PAN-3208


NicolasMassart:
@EdJoJob The thing is that this button is primarily expected to be used by occasional contributors, people who just want to fix a typo, so we can't expect them to know how to add the sign-off manually and it will inevitably lead to issues and more support. So I prefer to disable it given it never really proved its usefulness.

PR #40, opened 2019-10-01
[PAN-3023] Add doc for target-gas-limit command line option

sgregglives:

## PR description
Document the new target-gas-limit command line option.

Open questions: 

- What is the default value of the target-gas-limit command line option?

- Does it have a maximum value?

The Besu documentation has not been built and tested locally because I currently cannot run mkdocs.

## Fixed Issue(s)

PAN-3023


docs/Reference/CLI/CLI-Syntax.md
```diff
@@ -1419,6 +1419,27 @@ rpc-ws-port="6174"
...
 The default is 8546. Ports must be [exposed appropriately](../../HowTo/Find-and-Connect/Configuring-Ports.md).
 
+### target-gas-limit
+
+```bash tab="Syntax"
+--target-gas-limit=<INTEGER>
+```
+
+```bash tab="Command Line"
+--target-gas-limit=nnn
```
MadelineMurray: Rather than `nnn` this needs to be an actual example - the point of including is so the user can look at this and see a specific example of the format.  eg, when I tried to start a node with this CLI initially I assumed the value would be in hex but it's decimal.  An example would have saved me going down the wrong path `--target-gas-limit=nnn`

docs/Reference/CLI/CLI-Syntax.md
```diff
@@ -1422,6 +1422,28 @@ rpc-ws-port="6174"
...
+
+```bash tab="Environment Variable"
+BESU_TARGET_GAS_LIMIT=8000000
+```
+
+```bash tab="Configuration File"
+target-gas-limit="8000000"
+```
+
+Specifies the gas limit toward which Besu will gradually move on an existing network, if enough miners are in agreement. Use `target-gas-limit` to change the block gas limit set in the genesis file without creating a new network. The gas limit between blocks can change only 1/1024th, so the target tells the block creator how to set the gas limit in its block. If the values are the same or within 1/1024th, the limit is set to the specified value. Otherwise, the limit moves as far as it can within that constraint.
```
MadelineMurray: I think this is very close - just a few minor suggestions: 

1. if enough miners -> if enough miners/validators 

Explanation: Block creators are miners in PoW networks (ie, MainNet) and validators in PoA (ie, Clique or IBFT 2.0) networks 

2. to change the block gas limit set in the genesis file without -> to change the block gas limit without

Explanation: This sounded a like it was actually being updated in the genesis file but it's not -> the genesis file defines the value for the genesis (ie, first) block and isn't ever updated 

3. If the values are the same or within 1/1024th, the limit is set to the specified value. Otherwise, the limit moves as far as it can within that constraint. -> The block gas limit can move up to 1/1024th of the gas limit between blocks. The block creator moves the block gas limit as far as it can in the direction of the target gas limit. 

Explanation -> I think it's unlikely to be within the 1/1024th constraint so reworded to remove that part

docs/Reference/CLI/CLI-Syntax.md
```diff
@@ -1444,6 +1444,28 @@ rpc-ws-port="6174"
...
+BESU_TARGET_GAS_LIMIT=8000000
+```
+
+```bash tab="Configuration File"
+target-gas-limit="8000000"
+```
+
+Specifies the gas limit toward which Besu will gradually move on an existing network, if enough miners or validators are in agreement. Use `target-gas-limit` to change the block gas limit without creating a new network. The gas limit between blocks can change only 1/1024th, so the target tells the block creator how to set the gas limit in its block. The block gas limit can move up to 1/1024th of the gas limit between blocks. The block creator moves the block gas limit as far as it can in the direction of the target gas limit.
+    
+If a value for `target-gas-limit` is not specified, the block gas limit remains at the value specified in the [genesis file](../Config-Items.md#genesis-block-parameters).
```
shemnon:
```suggestion
If a value for `target-gas-limit` is not specified, the block gas limit is not changed between blocks.
```
We respect the current network value regardless of what the genesis file says (assuming it follows the rules in its movement)

docs/Reference/CLI/CLI-Syntax.md
```diff
@@ -1419,6 +1419,27 @@ rpc-ws-port="6174"
...
+```bash tab="Environment Variable"
+BESU_TARGET_GAS_LIMIT=nnn
+```
+
+```bash tab="Configuration File"
+target-gas-limit="nnn"
+```
+
+Specifies the gas limit toward which Besu will gradually move on an existing network. Use `target-gas-limit` to change the block gas limit without creating a new network. 
+The default is nnn.
```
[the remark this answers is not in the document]
sgregglives: Fixed.

docs/Reference/CLI/CLI-Syntax.md
```diff
@@ -1419,6 +1419,27 @@ rpc-ws-port="6174"
...
+
+```bash tab="Environment Variable"
+BESU_TARGET_GAS_LIMIT=nnn
+```
+
+```bash tab="Configuration File"
+target-gas-limit="nnn"
+```
+
+Specifies the gas limit toward which Besu will gradually move on an existing network. Use `target-gas-limit` to change the block gas limit without creating a new network.
```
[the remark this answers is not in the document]
sgregglives: Fixed.

docs/Reference/CLI/CLI-Syntax.md
sgregglives: Implemented these changes.

docs/Reference/CLI/CLI-Syntax.md
sgregglives: Added realistic example

PR #48, opened 2019-10-10
Bootnode updates

MadelineMurray:
Bootnode updates

Fixes PIE-1699


docs/HowTo/Find-and-Connect/Bootnodes.md
```diff
@@ -4,46 +4,24 @@ description: Configuring bootnodoes
...
+dropped as peers as the node continues to discover peers.
 
 ## Mainnet and Public Testnets
 
 For mainnet, Rinkeby, Ropsten, and Görli, Hyperledger Besu predefines a list of enode URLs.  
 
 ## Private Networks
 
-In private networks for development or testing purposes, specify one bootnode as described below.
+In private networks for development or testing purposes, specify at least one bootnode.
```
hmijail: ... unless one is going to use static nodes, in which case the whole discovery and bootnodes thing is useless.
I'd suggest it would be good to offer an initial panorama of the configuration alternatives: discovery+bootnodes vs static nodes in their different flavors.
This applies the same to production networks.

NicolasMassart: Given we are in the bootnodes page here, I think we can assume people already decided that they want to use bootnodes over static peers.

hmijail: My point is that, AFAICT, it is never explained that there is a decision to be made. Therefore I would expect users to just read sequentially the instructions and accumulate perceived-needed options as they go. That's what I would do, at least.

Note also that getting a half-explanation of the use cases and whether/when does it make sense to mix discovery and static nodes was kinda painful even among the developers who created and implemented the use cases. Retracing this information in the future will only get more difficult.

MadelineMurray: Have add a tip above this that indicates there is a decision to be made.

PR #50, opened 2019-10-10
[PAN-3205] Add note that states nat-method can't be used with Docker

sgregglives:

## PR description
`nat-method` currently can't be used on a Docker image, so a note was added.

## Fixed Issue(s)

PAN-3205

### DIFF
--- a/docs/Reference/CLI/CLI-Syntax.md
+++ b/docs/Reference/CLI/CLI-Syntax.md
@@ -578,19 +578,24 @@ The default is 1000.
 ### nat-method
 
 ```bash tab="Syntax"
---nat-method=<METHOD>
-```
-
-```bash tab="Command Line"
---nat-method=upnp
+--nat-method=UPNP
 ```
 
 ```bash tab="Example Configuration File"
-nat-method="upnp"
+nat-method="UPNP"
 ```
 
-Specifies the method for handling [NAT environments](../../HowTo/Find-and-Connect/Using-UPnP.md). 
-Options are `upnp` and `none`. The default is `none` (that is, NAT functionality is disabled).
+Specify the method for handling [NAT environments](../../HowTo/Find-and-Connect/Using-UPnP.md). Options are: `UPNP` and `NONE`.
+The default is `NONE`, which disables NAT functionality.
+
+!!!tip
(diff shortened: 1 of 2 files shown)


docs/Reference/CLI/CLI-Syntax.md
```diff
@@ -578,19 +578,24 @@ The default is 1000.
...
+nat-method="UPNP"
 ```
 
-Specifies the method for handling [NAT environments](../../HowTo/Find-and-Connect/Using-UPnP.md). 
-Options are `upnp` and `none`. The default is `none` (that is, NAT functionality is disabled).
+Specify the method for handling [NAT environments](../../HowTo/Find-and-Connect/Using-UPnP.md). Options are: `UPNP` and `NONE`.
+The default is `NONE`, which disables NAT functionality.
+
+!!!tip
+    `UPNP` works well with a typical home or small office environment where a wireless router or modem provides NAT isolation. This should provide
```
MadelineMurray: I think this tip could be updated to be the same as the one on the HowTo/Find-and-Connect/Using-UPnP page: 
UPnP support is often disabled by default in networking firmware. If disabled by default, explicitly enable UPnP support.

docs/Reference/CLI/CLI-Syntax.md
```diff
@@ -578,19 +578,24 @@ The default is 1000.
...
+    automatic detection and port-forwarding. UPnP support is often disabled by default in networking equipment firmware, however, any may need to be
+    explicitly enabled.
+
+!!!notes
+    * Option `UPNP` may introduce delays during node startup, especially on networks where no UPnP gateway device can be found.
+    * `--nat-method` cannot be used on a Docker image.
```
MadelineMurray: `on a Docker image` -> with the [Besu Docker image](link to the Run Besu from Docker Image page)

docs/Reference/CLI/CLI-Syntax.md
sgregglives: Fixed.

docs/Reference/CLI/CLI-Syntax.md
sgregglives: Fixed.

PR #54, opened 2019-10-23
[PIE-3110] Update postman collection.

bgravenorst:
Signed-off-by: Byron Gravenorst <byron.gravenorst@consensys.net>

## PR description
Update the Postman collection in Besu.

## Fixed Issue(s)
PIE-3110


NicolasMassart:
Could we have the collection file also uploaded in the doc site in /doc/postman or something so if the postman site is down for any reason, we are still able to get it?

bgravenorst:
When adding the link in a global file. The link is based on the position of the topic using the global file. So while `../../files/<blah>` might work in one topic, it 404's on another because their levels differ in the Nav structure. 

Also, instead of downloading the file, it displays the contents of the Postman JSON file on screen, which could confuse readers.

I think it is worthwhile doing this, but we probably need to first iron out these issues.

MadelineMurray:
I've raised a JI to add the file separately.

PR #56, opened 2019-10-23
[PAN-3252] Update MacOS requirements

sgregglives:

## PR description
MacOS requirements added.

## Fixed Issue(s)

PAN-3252


docs/HowTo/Get-Started/Build-From-Source.md
```diff
@@ -8,21 +8,24 @@ description: Building Hyperledger Besu from source code
 * [Java JDK](http://www.oracle.com/technetwork/java/javase/downloads/index.html)
 
 !!!important
```
sgregglives: This same note is labeled **Important** in this file and **Attention** in the other file. Which one is correct?

MadelineMurray: I'd use important - I don't think we use attention much at all and not sure the distinction is clear.

PR #65, opened 2019-11-05
[PAN-3049] Add CLI options for pruning

sgregglives:

## PR description
Update the CLI syntax reference with new options for pruning

Two flags are hidden in the CLI help. I assume we explain these types of hidden flags in the doc.

## Fixed Issue(s)

PAN-3049


docs/Reference/CLI/CLI-Syntax.md
```diff
@@ -1059,7 +1058,30 @@ BESU_PRIVACY_URL=http://127.0.0.1:8888
...
+
+```bash tab="Environment Variable"
+BESU_PRUNING_ENABLED=true
+```
+
+```bash tab="Configuration File"
+pruning-enabled=true
+```
+
+Enables pruning of unused world state of old blocks that will not be required except by chain reorganizations.
```
MadelineMurray: This relates a bit to my first comment on the conceptual material.  If I didn't already know what pruning was, I could read this and still be wondering why I would enable pruning.  

It won't work until the pruning concept material is added but it's probably easiest just to add the link now - they'll be merged at similar times.

RatanRSur: It's very important to note that the blocks are unlikely to required except by deep chain reorganizations. I also think it's not something that needs to be mentioned in the one-liner for what the option does. But definitely in the conceptual overview of pruning.

sgregglives: There was some discrepancy around this comment because the "outdated" label was missed, which became evident in a Slack discussion. I added the line about the pruned state trie nodes being not required to the [PR for the concept topic](https://github.com/hyperledger/besu-docs/pull/66) and changed chain to world state here.

docs/Reference/CLI/CLI-Syntax.md
```diff
@@ -1059,7 +1058,30 @@ BESU_PRIVACY_URL=http://127.0.0.1:8888
...
+
+```bash tab="Environment Variable"
+BESU_PRUNING_ENABLED=true
+```
+
+```bash tab="Configuration File"
+pruning-enabled=true
+```
+
+Enables [pruning](../../Concepts/Pruning.md) to reduce storage required for the world state size.
```
RatanRSur: I think we should remove `size` here. We're technically not worried about storing the size, we're worried about storing the thing itself :)

sgregglives: I believe the phrase came from a different comment during review. In any case, I removed the term.

PR #66, opened 2019-11-06
[PAN-3049] Add conceptual info about pruning

sgregglives:

## PR description
Update pruning info with a concept topic

## Fixed Issue(s)

PAN-3049

### DIFF
--- a/docs/Concepts/Privacy/Privacy-Overview.md
+++ b/docs/Concepts/Privacy/Privacy-Overview.md
@@ -9,6 +9,9 @@ Other participants cannot access the transaction content or list of participants
 !!! important
     For production systems requiring private transactions, we recommend using a network 
     with a consensus mechanism supporting transaction finality. For example, [IBFT 2.0](../../HowTo/Configure/Consensus-Protocols/IBFT.md).
+    
+!!! Caution
+    Do not use [pruning](../Pruning.md) if you plan to make private transactions.
 
 ## Private Transaction Manager
 
--- a/docs/Concepts/Privacy/Privacy-Overview.md
+++ b/docs/Concepts/Privacy/Privacy-Overview.md
@@ -11,7 +11,7 @@ Other participants cannot access the transaction content or list of participants
     with a consensus mechanism supporting transaction finality. For example, [IBFT 2.0](../../HowTo/Configure/Consensus-Protocols/IBFT.md).
     
 !!! Caution
-    Do not use [pruning](../Pruning.md) if you plan to make private transactions.
+    If you plan to use private transactions, do not use [pruning](../Pruning.md).
 
 ## Private Transaction Manager
 
--- a/docs/Concepts/Privacy/Privacy-Overview.md
+++ b/docs/Concepts/Privacy/Privacy-Overview.md
@@ -10,8 +10,8 @@ Other participants cannot access the transaction content or list of participants
     For production systems requiring private transactions, we recommend using a network 
     with a consensus mechanism supporting transaction finality. For example, [IBFT 2.0](../../HowTo/Configure/Consensus-Protocols/IBFT.md).
     
-!!! Caution
-    If you plan to use private transactions, do not use [pruning](../Pruning.md).
+!!! important
+    Using private transactions with [pruning](../Pruning.md) is not supported.
 
 ## Private Transaction Manager
 
--- a/docs/Concepts/Pruning.md
+++ b/docs/Concepts/Pruning.md
@@ -3,11 +3,11 @@ description: Pruning
 
 # Pruning
 
-Use pruning on Mainnet to remove state tries that aren't necessary. Otherwise, the state will continue to grow, and storage will become a problem.
+As the number of transactions increases, the world state grows, and storage can become a problem. Use pruning on any network to remove state tries that aren't necessary.
 
 !!! Caution
     Do not use pruning if you plan to make [private transactions](Privacy/Privacy-Overview.md).
 
-Before you enable pruning, ensure you have at least 600 GB of available space. To toggle pruning enablement, use the [`--pruning-enabled`](../Reference/CLI/CLI-Syntax.md#pruning-enabled) option.
+For a full sync to the Ethereum Mainnet with pruning enabled requires 600GB of disk space. To enable pruning, use the [`--pruning-enabled`](../Reference/CLI/CLI-Syntax.md#pruning-enabled) option.
 
 Pruning might increase block import times, but it does not affect the ability of nodes to stay in sync.
--- a/docs/Concepts/Pruning.md
+++ b/docs/Concepts/Pruning.md
@@ -3,11 +3,11 @@ description: Pruning
 
 # Pruning
 
-As the number of transactions increases, the world state grows, and storage can become a problem. Use pruning on any network to remove state tries that aren't necessary.
+Use pruning to reduce storage required for the chain. As the number of transactions increases, the world state grows. Pruning removes state tries that aren't required.
 
 !!! Caution
     If you plan to use [private transactions](Privacy/Privacy-Overview.md), do not use pruning.
 
-A full sync to the Ethereum Mainnet with pruning enabled requires 600GB of disk space. To enable pruning, use the [`--pruning-enabled`](../Reference/CLI/CLI-Syntax.md#pruning-enabled) option.
+A full sync to the Ethereum Mainnet with pruning enabled requires 750GB of disk space. To enable pruning, use the [`--pruning-enabled`](../Reference/CLI/CLI-Syntax.md#pruning-enabled) option.
 
 Pruning might increase block import times, but it does not affect the ability of nodes to stay in sync.
--- a/docs/Concepts/Pruning.md
+++ b/docs/Concepts/Pruning.md
@@ -3,11 +3,11 @@ description: Pruning
 
 # Pruning
 
-Use pruning to reduce storage required for the chain. As the number of transactions increases, the world state grows. Pruning removes state tries that aren't required.
+Use pruning to reduce storage required for the world state size. Pruning removes state trie nodes that aren't required.
 
-!!! Caution
-    If you plan to use [private transactions](Privacy/Privacy-Overview.md), do not use pruning.
+!!! Important
+    Using pruning with [private transactions](Privacy/Privacy-Overview.md) is not supported.
 
-A full sync to the Ethereum Mainnet with pruning enabled requires 750GB of disk space. To enable pruning, use the [`--pruning-enabled`](../Reference/CLI/CLI-Syntax.md#pruning-enabled) option.
+To run a full sync to the Ethereum Mainnet with pruning enabled, we recommend approximately 750GB of available disk space per node. To enable pruning, use the [`--pruning-enabled`](../Reference/CLI/CLI-Syntax.md#pruning-enabled) option.
 
 Pruning might increase block import times, but it does not affect the ability of nodes to stay in sync.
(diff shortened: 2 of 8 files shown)


docs/Concepts/Pruning.md
```diff
@@ -0,0 +1,13 @@
+description: Pruning
+<!--- END of page meta data -->
+
+# Pruning
+
+Use pruning on Mainnet to remove state tries that aren't necessary. Otherwise, the state will continue to grow, and storage will become a problem.
```
MadelineMurray: I think this needs flipping so the the first sentence states *why* you would use pruning and the second sentence describes *what* pruning is.

MadelineMurray: Also, you can use pruning on any network but it's most relevant on Mainnet because it has the largest world state.

docs/Concepts/Pruning.md
```diff
@@ -0,0 +1,13 @@
...
+
+!!! Caution
+    Do not use pruning if you plan to make [private transactions](Privacy/Privacy-Overview.md).
+
+Before you enable pruning, ensure you have at least 600 GB of available space. To toggle pruning enablement, use the [`--pruning-enabled`](../Reference/CLI/CLI-Syntax.md#pruning-enabled) option.
```
MadelineMurray: The first sentence isn't quite right for a couple of reasons: 
* The 600 GB of space relates to mainnet only.  ie, you could use pruning on Ropsten and the space requirement would be different. 
* Having enough space is a requirement for synchronizing the chain - the way this is worded makes it sound like space is required because of pruning which is misleading because the point of pruning is to reduce the storage required.

MadelineMurray: To toggle pruning enablement -> To enable pruning 

Suggested edit to both be more active and consistent with the rest of the doc

docs/Concepts/Privacy/Privacy-Overview.md
```diff
@@ -9,6 +9,9 @@ Other participants cannot access the transaction content or list of participants
 !!! important
     For production systems requiring private transactions, we recommend using a network 
     with a consensus mechanism supporting transaction finality. For example, [IBFT 2.0](../../HowTo/Configure/Consensus-Protocols/IBFT.md).
+    
+!!! Caution
+    Do not use [pruning](../Pruning.md) if you plan to make private transactions.
```
MadelineMurray: I think we've talked about this before in a sync but to reduce the amount users have to read put the _if_ part of sentences first. Then if that part of the sentence doesn't apply to them, they don't need to read the dependant piece. eg, 

If make plan to use private transactions, do not use pruning. 

make -> use.  This constraint on pruning and private transactions applies when receiving private transactions as well.

docs/Concepts/Pruning.md
```diff
@@ -0,0 +1,13 @@
+description: Pruning
+<!--- END of page meta data -->
+
+# Pruning
+
+As the number of transactions increases, the world state grows, and storage can become a problem. Use pruning on any network to remove state tries that aren't necessary.
```
MadelineMurray: Suggested change: 
Use pruning to reduce storage required for the chain.  As the number of transaction increases, the world state grows. Pruning removes state tries that aren't required.

docs/Concepts/Privacy/Privacy-Overview.md
sgregglives: Updated

docs/Concepts/Pruning.md
sgregglives: Updated

docs/Concepts/Pruning.md
sgregglives: Updated

docs/Concepts/Privacy/Privacy-Overview.md
```diff
@@ -9,6 +9,9 @@ Other participants cannot access the transaction content or list of participants
 !!! important
     For production systems requiring private transactions, we recommend using a network 
     with a consensus mechanism supporting transaction finality. For example, [IBFT 2.0](../../HowTo/Configure/Consensus-Protocols/IBFT.md).
+    
+!!! Caution
+    If you plan to use private transactions, do not use [pruning](../Pruning.md).
```
RatanRSur: Perhaps we can say that pruning isn't supported with private transactions? The current wording makes it sound like we don't enforce it and they're in charge of it.

docs/Concepts/Pruning.md
```diff
@@ -0,0 +1,13 @@
...
+<!--- END of page meta data -->
+
+# Pruning
+
+Use pruning to reduce storage required for the chain. As the number of transactions increases, the world state grows. Pruning removes state tries that aren't required.
+
+!!! Caution
+    If you plan to use [private transactions](Privacy/Privacy-Overview.md), do not use pruning.
+
+A full sync to the Ethereum Mainnet with pruning enabled requires 750GB of disk space. To enable pruning, use the [`--pruning-enabled`](../Reference/CLI/CLI-Syntax.md#pruning-enabled) option.
```
RatanRSur: Perhaps we should be more specific here, it doesn't require 750GB of space *now*, but rather, 750GB is a recommendation to allow for chain state growth over the course of months. Maybe something like: "We recommend that a node be allocated 750GB of disk space to accommodate Ethereum state growth until approximately May 2020 or 600GB until February 2020."

docs/Concepts/Pruning.md
```diff
@@ -0,0 +1,13 @@
+description: Pruning
+<!--- END of page meta data -->
+
+# Pruning
+
+Use pruning to reduce storage required for the chain. As the number of transactions increases, the world state grows. Pruning removes state tries that aren't required.
+
+!!! Caution
+    If you plan to use [private transactions](Privacy/Privacy-Overview.md), do not use pruning.
```
RatanRSur: same as above

docs/Concepts/Pruning.md
```diff
@@ -0,0 +1,13 @@
+description: Pruning
+<!--- END of page meta data -->
+
+# Pruning
+
+Use pruning to reduce storage required for the chain. As the number of transactions increases, the world state grows. Pruning removes state tries that aren't required.
```
RatanRSur: `chain` -> `world-state size`

The world state isn't guaranteed to grow along with transactions so I would just remove that sentence.

`state tries` -> `state-trie nodes`

docs/HowTo/Get-Started/System-Requirements.md
```diff
@@ -22,8 +22,8 @@ A sample [Grafana dashboard](https://grafana.com/grafana/dashboards/10273) is pr
 
 ## RAM 
 
-Besu requires 4GB of RAM. For Ethereum Mainnet, a minimum of 8GB of RAM is required 
+Besu requires 4GB of RAM. For Ethereum Mainnet, a minimum of 8GB of RAM is required. 
 
 ## Disk Space 
 
-Syncing to the Ethereum mainnet requires 3TB for a full sync. 
+Syncing to the Ethereum Mainnet requires 3TB for a full sync. To reduce the requirement, use [pruning](../../Concepts/Pruning.md). With pruning, you need 750GB.
```
RatanRSur: "we recommend"

docs/Reference/CLI/CLI-Syntax.md
```diff
@@ -986,8 +986,10 @@ BESU_PRIVACY_ENABLED=false
 privacy-enabled=false
 ```
 
-Set to enable [private transactions](../../Concepts/Privacy/Privacy-Overview.md). 
-The default is false.  
+Set to enable [private transactions](../../Concepts/Privacy/Privacy-Overview.md). The default is false.
+
+!!! Caution
+    Do not use [pruning](../../Concepts/Pruning.md) if you plan to make private transactions.
```
RatanRSur: same as above

docs/Concepts/Pruning.md
sgregglives: This is updated but with some additional changes from the tech review.

docs/Concepts/Privacy/Privacy-Overview.md
sgregglives: Updated.

docs/Concepts/Pruning.md
sgregglives: Per our conversation in Slack, I've updated this section to include just the recommendation for 750GB.

docs/Concepts/Pruning.md
sgregglives: Updated.

docs/Concepts/Pruning.md
sgregglives: Updated.

docs/HowTo/Get-Started/System-Requirements.md
sgregglives: Updated.

docs/Reference/CLI/CLI-Syntax.md

docs/Concepts/Pruning.md
```diff
@@ -3,11 +3,11 @@ description: Pruning
 
 # Pruning
 
-Use pruning to reduce storage required for the chain. As the number of transactions increases, the world state grows. Pruning removes state tries that aren't required.
+Use pruning to reduce storage required for the world state size. Pruning removes state trie nodes that aren't required.
```
RatanRSur: Same comment as the other PR about `size`

docs/Concepts/Pruning.md
```diff
@@ -3,11 +3,11 @@ description: Pruning
...
 
-!!! Caution
-    If you plan to use [private transactions](Privacy/Privacy-Overview.md), do not use pruning.
+!!! Important
+    Using pruning with [private transactions](Privacy/Privacy-Overview.md) is not supported.
 
-A full sync to the Ethereum Mainnet with pruning enabled requires 750GB of disk space. To enable pruning, use the [`--pruning-enabled`](../Reference/CLI/CLI-Syntax.md#pruning-enabled) option.
+To run a full sync to the Ethereum Mainnet with pruning enabled, we recommend approximately 750GB of available disk space per node. To enable pruning, use the [`--pruning-enabled`](../Reference/CLI/CLI-Syntax.md#pruning-enabled) option.
```
RatanRSur: Was the `per node` added because it's common for our users to have multiple nodes using the same storage? If not, I think it's clearer to remove that caveat.

docs/HowTo/Get-Started/System-Requirements.md
```diff
@@ -26,4 +26,4 @@ Besu requires 4GB of RAM. For Ethereum Mainnet, a minimum of 8GB of RAM is requi
 
 ## Disk Space 
 
-Syncing to the Ethereum Mainnet requires 3TB for a full sync. To reduce the requirement, use [pruning](../../Concepts/Pruning.md). With pruning, you need 750GB.
+Syncing to the Ethereum Mainnet requires 3TB for a full sync. To reduce the requirement, use [pruning](../../Concepts/Pruning.md). With pruning, we recommend approximately 750GB of available disk space per node.
```
RatanRSur: Same `per node` comment

docs/Concepts/Pruning.md
sgregglives: Removed the term.

docs/Concepts/Pruning.md
sgregglives: Updated.

docs/HowTo/Get-Started/System-Requirements.md
sgregglives: Updated.

PR #69, opened 2019-11-07
[PIE-2076] Add account with nonzero balance to examples of eth_getBalance

sgregglives:

## PR description
Add account with a balance to the examples.

Values in the Results tabs will not match output from tests.

## Fixed Issue(s)

PIE-2076

### DIFF
--- a/docs/Reference/API-Methods.md
+++ b/docs/Reference/API-Methods.md
@@ -4,10 +4,9 @@ description: Hyperledger Besu JSON-RPC API methods reference
 # Hyperledger Besu API Methods
 
 !!! attention
-    All JSON-RPC HTTP examples use the default host and port endpoint `http://127.0.0.1:8545`. 
-
-    If using the [--rpc-http-host](CLI/CLI-Syntax.md#rpc-http-host) or [--rpc-http-port](CLI/CLI-Syntax.md#rpc-http-port)
-    options, update the endpoint.  
+    * All JSON-RPC HTTP examples use the default host and port endpoint `http://127.0.0.1:8545`. If using the [--rpc-http-host](CLI/CLI-Syntax.md#rpc-http-host) or [--rpc-http-port](CLI/CLI-Syntax.md#rpc-http-port)
+    options, update the endpoint.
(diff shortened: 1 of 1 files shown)


docs/Reference/API-Methods.md
```diff
@@ -887,22 +887,25 @@ Returns the account balance of the specified address.
 
 **Returns**
 
-`result` : *QUANTITY* - Integer value of the current balance in wei.
+`result` : *QUANTITY* - Current balance in wei as a hexadecimal value.
+
+!!! important
```
MadelineMurray: This note is correct but what it's saying applies to all examples.  I think we need to remove this one and add something general along these lines to the Attention box at the top of this page.  Something like: 

Except for the examples made on the Ropsten network, the example requests are made against private networks and your example results may be different depending on network configuration and activity.

sgregglives: Updated.

docs/Reference/API-Methods.md
```diff
@@ -4,10 +4,9 @@ description: Hyperledger Besu JSON-RPC API methods reference
 # Hyperledger Besu API Methods
 
 !!! attention
-    All JSON-RPC HTTP examples use the default host and port endpoint `http://127.0.0.1:8545`. 
-
-    If using the [--rpc-http-host](CLI/CLI-Syntax.md#rpc-http-host) or [--rpc-http-port](CLI/CLI-Syntax.md#rpc-http-port)
-    options, update the endpoint.  
+    * All JSON-RPC HTTP examples use the default host and port endpoint `http://127.0.0.1:8545`. If using the [--rpc-http-host](CLI/CLI-Syntax.md#rpc-http-host) or [--rpc-http-port](CLI/CLI-Syntax.md#rpc-http-port)
+    options, update the endpoint.
+    * Except for the examples made on the Ropsten network, the example requests are made against private networks, and your example results may be different, depending on network configuration and activity.
```
MadelineMurray: Suggest breaking the sentence in two so it's easier to read: 

Except for the examples made on the Ropsten network, the example requests are made against private networks. Depending on network configuration and activity, your example results may be different.

sgregglives: Updated

PR #70, opened 2019-11-07
[PIE-2081] Update min-gas-price to be clear that it's the same value returned by eth_gasPrice

sgregglives:
## PR description
Add information about the relationship between the value of min-gas-price and the value returned by the eth_gasPrice method

## Fixed Issue(s)

PIE-2081

### DIFF
--- a/docs/Reference/API-Methods.md
+++ b/docs/Reference/API-Methods.md
@@ -743,7 +743,7 @@ None
 
 ### eth_gasPrice
 
-Returns the current gas unit price in a running node in wei. It is the hexadecimal equivalent of the price specified by the [`--min-gas-price`](CLI/CLI-Syntax.md#min-gas-price) command line option when the node was started.
+Returns the current gas unit price in a running node in wei. It is the hexadecimal equivalent of the price specified by the [`--min-gas-price`](CLI/CLI-Syntax.md#min-gas-price) command line option when the node was started or the default.
 
 **Parameters**
 
--- a/docs/Reference/API-Methods.md
+++ b/docs/Reference/API-Methods.md
@@ -743,7 +743,7 @@ None
 
 ### eth_gasPrice
 
-Returns the current gas unit price in a running node in wei. It is the hexadecimal equivalent of the price specified for the [`--min-gas-price`](CLI/CLI-Syntax.md#min-gas-price) command line option when the node was started or the default.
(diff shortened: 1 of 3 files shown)


docs/Reference/CLI/CLI-Syntax.md
```diff
@@ -593,7 +593,7 @@ BESU_MIN_GAS_PRICE=1337
 min-gas-price=1337
 ```
 
-The minimum price that a transaction offers for it to be included in a mined block.
+The minimum price that a transaction offers for it to be included in a mined block. It is the same as the value returned by [`eth_gasPrice`](../API-Methods.md).
```
MadelineMurray: Suggest updating this to reflect why you'd use `eth_gasPrice`. eg, 

Use `eth_gasPrice` to obtain the minimum gas price in a running node.  

ie, you use the `min-gas-price` CLI option to set the minimum gas price on startup but you can use `eth_gasPrice` to find out what it is in a running node.

sgregglives: I added some of this language to the min-gas-price CLI option description and the eth_gasPrice method description.

docs/Reference/API-Methods.md
```diff
@@ -743,7 +743,7 @@ None
 
 ### eth_gasPrice
 
-Returns the current gas unit price in wei.
+Returns the current gas unit price in a running node in wei. It is the hexadecimal equivalent of the price specified for the [`--min-gas-price`](CLI/CLI-Syntax.md#min-gas-price) command line option when the node was started or the default.
```
MadelineMurray: Returns the current gas unit price in a running node in wei -> Returns the current gas unit price in wei

_running node_ doesn't make sense there because the node has to be running in order to make a JSON-RPC request. It's relevant in the min-gas-price description because you can't see min-gas-price in a running node. 

the default -> the default minimum gas price 

It's not clear what default this is currently

docs/Reference/CLI/CLI-Syntax.md
```diff
@@ -593,7 +593,7 @@ BESU_MIN_GAS_PRICE=1337
 min-gas-price=1337
 ```
 
-The minimum price that a transaction offers for it to be included in a mined block.
+The minimum price that a transaction offers for it to be included in a mined block. To retrieve the minimum price in a running node later, use [`eth_gasPrice`](../API-Methods.md#eth_gasprice).
```
MadelineMurray: running node later, use -> running node, use

I don't think _later_ is needed here

docs/Reference/API-Methods.md
sgregglives: Updated.

docs/Reference/CLI/CLI-Syntax.md

PR #99, opened 2019-11-22
Add brief transaction pool definition

demmojo:
Signed-off-by: Mohamed Abdulaziz <mo@blok-z.com>

## PR description

Added a brief explanation about transaction pools.


docs/Concepts/Transactions/Transaction-Pool.md
```diff
@@ -1,5 +1,7 @@
 # Transaction Pool 
 
+Mining nodes maintain a transaction pool where pending transactions are stored before they are processed.
```
ajsutton: Note that all nodes contain a transaction pool, not just mining nodes.  It's an important part of ensuring transactions are propagated around the network.

demmojo: Thanks for clearing that up for me! Changed the text to 'all nodes' rather than just 'mining nodes'.

PR #123, opened 2020-01-20
Added default data location 

MadelineMurray:

Added default data location after getting user feedback that it wasn't clear where the data was if the --data-path option wasn't specified (ie, the default behaviour).


docs/HowTo/Backup/Backup.md
```diff
@@ -13,8 +13,11 @@ stored under source control.
 
 ## Data Backups
 
-We recommend mounting a separate volume to store data. Use the [`--data-path`](../../Reference/CLI/CLI-Syntax.md#data-path) 
-command line option to pass the path to Besu. 
+We recommend mounting a [separate volume to store data](../Get-Started/Run-Docker-Image.md#starting-besu). 
+Use the [`--data-path`](../../Reference/CLI/CLI-Syntax.md#data-path) command line option to pass the path to Besu. 
+
+If installed locally, the default data location is the directory in which Besu is installed. To
```
EdJoJob: It feels to me that these two paragraphs can be combined, with then only a single reference to the `--data-path` being used to specify the path

joshuafernandes: Can we put `The default data location is the directory in which Besu is installed.` before the `We recommend ....` and then can leave that last part of specifying the path out?

MadelineMurray: Fixed

PR #132, opened 2020-01-31
Added content on protocol upgrades

MadelineMurray:
Added content on protocol upgrades


docs/Concepts/Protocol-Upgrades.md
```diff
@@ -0,0 +1,25 @@
+description: Protocol Upgrades      
+<!--- END of page meta data -->
+
+# Protocol upgrades in private networks 
+
+Hardforks are the mechanism for upgrading the Ethereum protocol.
```
shemnon: Style:  All Core Devs is moving towards "Network Upgrades" instead of "Hardforks."  Not sure how that figures into this doc...  I like Protocol Upgrade and Network Upgrade is the point in time the Protocol Upgrade is applied.  I think they are trying to reserve Fork for events like the ETC or BCH split.

MadelineMurray: Removed references to hardforks and included distinction about network upgrade being the point in time as which the protocol upgrade is applied.

PR #165, opened 2020-02-13
updated CI and customised checks

### LINKED ISSUE #160: Customized markdownlint style file required to override mdl --style all
To override the default rules provided by markdownlint we need a style file and update the `--style=all` option in the mdl command in circleci to use this new style.

https://consensys.slack.com/archives/GCEHWHAQ2/p1581509902077000

NicolasMassart:
## PR description

I’m pushing a PR to have a better CI (learned from our other projects best practices) on Besu doc. It Includes a change on the line length rule to match 100 and the title punctuation modified rule. Plus the scripts to run the test locally by yourself. I added a readme in the CI/scripts directory.

- updated CI config to remove useless DIND -> now directly using the target
image with the correct soft installed as it was otherwise not possible to run
the jobs locally with Circle CI CLI.
- Updated and customised Vale config rules
- Updated en customised markdownlint config and rules
- moved config files to CI folder
- added scripts to run CI jobs locally with Circle CI CLI.

## Fixed Issue(s)
fixes #160


CI/markdownlint/style.rb
```diff
@@ -0,0 +1,8 @@
+all
+
+rule "no-duplicate-header", :allow_different_nesting => true
+rule 'line-length', :line_length=>100, :code_blocks=>false, :tables=> false
+rule 'no-trailing-punctuation', :punctuation=>'.,;:!'
```
grantnoble: In `CONTRIBUTING.md` we have a header "Thanks for taking the time to contribute!", so perhaps remove "!" from the punctuation list.

NicolasMassart: Will do in a content changing PR. This one is really only for the rules to be set.

PR #180, opened 2020-02-18
mdlint and vale editorial updates in Concepts/Consensus-Protocols directory

grantnoble:
Line spacing <= 100
Remove end of line spaces
Passive voice -> active voice
Misc wording changes


docs/Concepts/Consensus-Protocols/Overview-Consensus.md
```diff
@@ -1,50 +1,48 @@
 description: Besu consensus protocols
 <!--- END of page meta data -->
 
-# Consensus Protocols 
+# Consensus protocols
 
-Besu implements a number of consensus protocols: 
+Besu implements a number of consensus protocols, including:
```
MadelineMurray: I'm not sure if this is just me but does adding _including_ here make it sound like the consensus protocols listed here are a subset of consensus protocols implemented in Besu rather than being all of them?

grantnoble: Updated.

PR #210, opened 2020-02-26
Added auto log bloom caching option

MadelineMurray:

Added auto log bloom caching option


docs/Reference/CLI/CLI-Syntax.md
```diff
@@ -33,6 +33,28 @@ To start a Besu node run:
...
+```bash tab="Environment Variable"
+BESU_AUTO_LOG_BLOOM_CACHING_ENABLED=false
+```
+
+```bash tab="Example Configuration File"
+auto-log-bloom-caching-enabled=false
+```
+
+Enables or disables automatic log bloom caching. APIs such as [`eth_getLogs`](../API-Methods.md#eth_getlogs)
+and [`eth_getFilterLogs`](../API-Methods.md#eth_getfilterlogs) use the indexes for improved performance.
```
abdelhamidbakhta: Maybe `cache` instead of `indexes`

MadelineMurray: Done

docs/Reference/CLI/CLI-Syntax.md
```diff
@@ -33,6 +33,28 @@ To start a Besu node run:
 ```bash
 besu [OPTIONS] [COMMAND]
 ```
+
+### auto-log-bloom-caching-enabled
+
+```bash tab="Syntax"
+--auto-log-bloom-caching-enabled=false
```
NicolasMassart: I feel like the syntax should be something like `--auto-log-bloom-caching-enabled=<true|false>` and the examples have one of the two values only like for instance for https://besu.hyperledger.org/en/stable/Reference/CLI/CLI-Syntax/#p2p-enabled but it seems that boolean options are not all displayed the same way in the doc, so maybe we'll make a consistency pass later.

MadelineMurray: @NicolasMassart - agree we should fix this/make them consistent and I'm not going to do that as part of this PR. Could you raise an issue so we remember to do this?

Issue #206, opened 2020-02-25
Homebrew install: Error: Formulae found in multiple taps


NicolasMassart:
User feedback on HotJar about brew install doc:

https://besu.hyperledger.org/en/latest/HowTo/Get-Started/Install-Binaries/

> When I follow the instructions for installing via brew, I get this error. Easy fix, but the docs are wrong:
> 
> Error: Formulae found in multiple taps:
> * hyperledger/besu/besu
> * pegasyseng/pantheon/besu
> 
> Please use the fully-qualified name (e.g. hyperledger/besu/besu) to refer to the formula.

PR #213, opened 2020-02-27
Updated brew install instructions per user feedback

MadelineMurray:
Updated brew install instructions per user feedback in issue #206

### DIFF
--- a/docs/HowTo/Get-Started/Install-Binaries.md
+++ b/docs/HowTo/Get-Started/Install-Binaries.md
@@ -20,7 +20,7 @@ description: Install Hyperledger Besu from binary distribution
 
 ```bash
 brew tap hyperledger/besu
-brew install besu
+brew install hyperledger/besu/besu
 ```
 Display Besu command line help to confirm installation:
 


ajsutton:
With the 1.4 release we've removed the old PegaSysEng/pantheon/brew tap and migrated it over to point to hyperledger/besu so I don't this this change is required anymore.  Probably doesn't hurt though.

Would be really useful to include a summary of the user feedback in the PR description for historical reference as well.

MadelineMurray:
@ajsutton - ironic timing in that case.  Linked the issue with the feedback.

ajsutton:
Yep so reviewing that feedback it should rapidly become a non-issue which is nice.

MadelineMurray:
@ajsutton - given it'll become a non-issue I'm inclined to not merge this one.  `brew install besu` is easier than `brew install hyperledger/besu/besu`. WDYT?

ajsutton:
I think the safe thing is to merge it. I think it happens when you have two taps that both provide besu.  That used to be PegaSysEng/pantheon (which provided a besu as part of the migration process that installed the same thing as hyperledger/besu).  But it could potentially be any tap the user added.

And I can't reproduce the original situation now so it's kind of hard to know for sure what situations someone's brew could wind up in depending on when they added and updated what.

So probably merge it anyway I guess...

MadelineMurray:
Thanks - makes sense given we can't control which besu taps might be added in future.

Issue #207, opened 2020-02-25
Warning needed on configuration starting a miner and opening RPC


NicolasMassart:
User feedback about configuration starting a miner and opening RPC.

https://besu.hyperledger.org/en/latest/HowTo/Get-Started/Starting-node/

> The configuration example on this page includes two things that seem ill-advised. There should, at least, be a warning to the user about (a) that this configuration starts a miner, and (b) that the RPC is 'wide open'. In the past, people said about both geth and parity that opening one's RPC in this way was asking for trouble. Maybe you shouldn't make that the default without at least a warning to the reader.

grantnoble:
PR #228 fixes this issue. Closing.

PR #228, opened 2020-03-05
Caution added after config for running a node for testing

grantnoble:
## PR description
In /HowTo/Get-Started/Starting-node.md, caution added after config for running a node for testing.
## Fixed Issue(s)
#207


docs/HowTo/Get-Started/Starting-node.md
```diff
@@ -94,6 +94,18 @@ rpc-http-enabled=true
 data-path="/tmp/tmpdata-path"
 ```
 
+!!! caution
+
+    Enabling the HTTP JSON-RPC service
```
NicolasMassart: maybe displayed on multiple lines or with an unordered list?

grantnoble: I will update accordingly.

docs/HowTo/Get-Started/Starting-node.md
```diff
@@ -94,6 +94,20 @@ rpc-http-enabled=true
 data-path="/tmp/tmpdata-path"
 ```
 
+!!! caution
+
+    We do not recommend using the following settings in a production environment:
```
NicolasMassart: Instead of "we" having the sentence like "There's a security risk to use the following settings in a production environment"

MadelineMurray: Using "we" is actually a standard we've taken from the Microsoft MoS (our default standard) - https://docs.microsoft.com/en-us/style-guide/a-z-word-list-term-collections/r/recommend 

Having said that, now we're open source I'm not sure that still makes sense.  Who is 'we'? 

@NicolasMassart

docs/HowTo/Get-Started/Starting-node.md
```diff
@@ -94,6 +94,20 @@ rpc-http-enabled=true
...
+    
+    * Enabling the HTTP JSON-RPC service
+      ([`--rpc-http-enabled`](../../Reference/CLI/CLI-Syntax.md#rpc-http-enabled)) and setting
+      [`-rpc-http-host`](../../Reference/CLI/CLI-Syntax.md#rpc-http-host) to 0.0.0.0 exposes the
```
MadelineMurray: exposes -> to expose 

Suggested edit to make parallel with the other two bullets.

docs/HowTo/Get-Started/Starting-node.md
grantnoble: Yes, quite a few instances of "We" in my travels through the Besu docs. I must admit I've never been a fan of using "we" for the reason you state, @MadelineMurray. Who is "we"?
Perhaps, "The following settings pose a security risk if used in production environments:"?

If I can raise another question about these statements? I've been using "production environment", but I've also seen "production code", and other variations on the same thing. Is there a phrase we want to standardise on?

NicolasMassart: On the second question: production environment is the term. There's no production code, the code, once released, is the same whatever you use it for. There's a non released code too, but our should talk only about the released one IMO.

grantnoble: Changed to "The following settings are a security risk in production environments:"

PR #230, opened 2020-03-08
Editorial mdlint and vale updates for docs/Reference directory

grantnoble:
Line length, sentence case for headings, EOL spaces, passive voice to active voice, other misc. editorial changes.


grantnoble:
There are quite a few instances of the Markdown lint warming "MD036 Emphasis used instead of a header". Where we use \*\*Parameters** and \*\*Returns**, for example.

NicolasMassart:
> There are quite a few instances of the Markdown lint warming "MD036 Emphasis used instead of a header". Where we use **Parameters** and **Returns**, for example.

I think we should switch to headers indeed. It's easier to link and navigate.

NicolasMassart:
Trailing spaces may also be simple to fix and would remove a lot of instance of this issue

grantnoble:
Agree about the headers. I will update.

docs/Reference/API-Methods.md
```diff
@@ -4256,50 +4490,62 @@ per transaction in the order the transactions were executed.
...
 
 !!! important
-    For production systems requiring private transactions, we recommend using a network
-    with a consensus mechanism supporting transaction finality. For example, [IBFT 2.0](../HowTo/Configure/Consensus-Protocols/IBFT.md).
-    
-    Using private transactions with [pruning](../Concepts/Pruning.md) or [fast sync](CLI/CLI-Syntax.md#sync-mode)
-    is not supported.
 
-    Besu does not implement [`eea_sendTransaction`](../HowTo/Send-Transactions/Account-Management.md).
+    For production systems requiring private transactions, we recommend using a network with a
```
NicolasMassart: According to https://docs.microsoft.com/en-us/style-guide/a-z-word-list-term-collections/w/we we should try another way write this sentence as here it doesn't explain the benefits of doing so.
So perhaps "For production systems requiring private transactions, use a network with a consensus mechanism supporting transaction finality to make sure the private state can not get inconsistant with the chain. For example, IBFT2 provides the required finality."

grantnoble: Reworded slightly, but updated accordingly.

PR #234, opened 2020-03-11
Editorial mdlint and vale edits to the Tutorials/Examples directory

grantnoble:
Line length, eol spaces, appropriate line breaks, passive voice to active voice, plus some heading changes with the corresponding links changed in additional files


grantnoble:
One lint warning I don't think we can do anything about: "Private-Network-Example.md:329: MD024 Multiple headers with the same content". Two "Prerequisite" headings in the same file.

PR #236, opened 2020-03-12
Added simple onchain privacy group example

MadelineMurray:

Added simple onchain privacy group example

### DIFF
--- a/docs/HowTo/Use-Privacy/Use-OnChainPrivacy.md
+++ b/docs/HowTo/Use-Privacy/Use-OnChainPrivacy.md
@@ -34,10 +34,8 @@ To create and find an onchain privacy group using the [web3.js-eea library](http
     node simpleOnChainPrivacy.js
     ```
 
-    Two transactions are created and mined. The first transaction initiates the onchain privacy group
-    and the second transaction adds the members. 
-
-    `findPrivacyGroup` finds and displays the privacy group created by the two transactions. 
+    The onchain privacy group is created with two members. `findPrivacyGroup` finds and displays the 
+    created privacy group. 
 
 !!! tip 
     In the Orion logs for Orion 1 and Orion 2, `PrivacyGroupNotFound` errors are displayed. The errors 
--- a/docs/HowTo/Use-Privacy/Use-OnChainPrivacy.md
+++ b/docs/HowTo/Use-Privacy/Use-OnChainPrivacy.md
@@ -20,7 +20,11 @@ membership of onchain privacy groups.
 ## Enabling onchain privacy groups 
 
 Use the [`--privacy-onchain-groups-enabled`](../../Reference/CLI/CLI-Syntax.md#privacy-onchain-groups-enabled)
-command line option to enable onchain privacy groups. 
+command line option to enable onchain privacy groups. When onchain privacy groups are enabled, the 
+[`priv_createPrivacyGroup`](../../Reference/API-Methods.md#priv_createprivacygroup), 
+[`priv_deletePrivacyGroup`](../../Reference/API-Methods.md#priv_deleteprivacygroup), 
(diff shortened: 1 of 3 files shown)


docs/HowTo/Use-Privacy/Use-OnChainPrivacy.md
```diff
@@ -0,0 +1,55 @@
...
+
+2. Run: 
+
+    ```
+    node simpleOnChainPrivacy.js
+    ```
+
+    Two transactions are created and mined. The first transaction initiates the onchain privacy group
+    and the second transaction adds the members. 
+
```
antonydenyer: I think it's only 1 transaction. This initiates a privacy group with 2 memebers.

docs/HowTo/Use-Privacy/Use-OnChainPrivacy.md
```diff
@@ -0,0 +1,55 @@
...
+## Adding and removing members 
+
+To add and remove members from an onchain privacy group, use the `addToPrivacyGroup` and 
+`removeFromPrivayGroup` methods in the [web3.js-eea library](https://github.com/PegaSysEng/web3js-eea)
+client library. 
+
+!!! note 
+    When a member is added, all existing group transactions are pushed to the new member. If there are a 
+    large number of existing transactions, pushing the transactions may take some time.
```
antonydenyer: The node being added will not process any other transactions until this process has finished.

iikirilov: ignore tony here

docs/HowTo/Use-Privacy/Use-OnChainPrivacy.md
```diff
@@ -0,0 +1,55 @@
...
+    path from onchain privacy groups created using v1.4 to enable use of onchain privacy group functionality 
+    in future versions. 
+
+    We do not recommend creating onchain privacy groups in a chain with existing 
+    [offchain privacy groups](../../Concepts/Privacy/Privacy-Groups.md). 
+    
+## Enabling onchain privacy groups 
+
+Use the [`--privacy-onchain-groups-enabled`](../../Reference/CLI/CLI-Syntax.md#privacy-onchain-groups-enabled)
+command line option to enable onchain privacy groups.
```
iikirilov: I don't think we actually have this

docs/HowTo/Use-Privacy/Use-OnChainPrivacy.md
antonydenyer: I should have been clearer, it will block until it's rehydrated the state. If you've got 1m transactions in a group and you add a node. The node being added will need to process them all before processing the next transaction, right?

docs/HowTo/Use-Privacy/Use-OnChainPrivacy.md
MadelineMurray: I was misreading the output here.  Updating.

docs/HowTo/Use-Privacy/Use-OnChainPrivacy.md
iikirilov: Yes - but that happens after the blob has been built and distributed.
The processing happens once the person being added has received the marker transaction which points to the private add payload.

It can be clarified that both the "pushing" and "processing" may take some time or you could change "pushing the transaction" to "adding a new a member" so it would read "adding a new a member may take some time"

MadelineMurray: Thanks for clarifying.  Have updated.

MadelineMurray:
> Should we mention that when onchain Privacy groups is enabled the other privacy group API methods are disabled?

Added this

PR #242, opened 2020-03-13
pruning-block-confirmations and pruning-blocks-retained options documented

grantnoble:
Documented as per https://github.com/hyperledger/besu/pull/440


docs/Reference/CLI/CLI-Syntax.md
```diff
@@ -1346,6 +1346,66 @@ privacy-url="http://127.0.0.1:8888"
...
+
+```bash tab="Environment Variable"
+BESU_PRUNING_BLOCK_CONFIRMATIONS=5
+```
+
+```bash tab="Configuration File"
+pruning-block-confirmations=5
+```
+
+The minimum number of confirmations on a block before marking begins. The default is 10.
```
NicolasMassart: what is marking?

grantnoble: Spoke with Abdel about this and have updated the description.

PR #275, opened 2020-03-24
Document the Java Virtual Machine requirements

### LINKED ISSUE #167: Add recommendation on size of Java VM
Add the recommended size of the Java VM to the docs:
Min 4GB for private. 16GB for mainnet, but really depends on your usage, network setup, etc.

Besu will use the most memory while syncing and then drop back.  Use monitoring to see what your node is doing.

@shemnon @joshuafernandes @EdJoJob are the input/reviewers for this.

grantnoble:
Fixes #167


docs/HowTo/Get-Started/System-Requirements.md
```diff
@@ -30,6 +30,24 @@ To determine system requirements, check CPU and disk space requirements using
 Besu requires 4GB of RAM. For public Ethereum networks, including MainNet and testnets such as
```
NicolasMassart: You should either remove or modify this sentence that contradicts your new text
Also why do you have this change also in #276 ?

grantnoble: Aren't RAM requirements and JVM memory requirements different things? Perhaps @EdJoJob you can help clarify?

EdJoJob: The RAM requirement should be the same as the JVM memory requirement, they are the same thing.  However people may interpret the RAM as what is available on the physical machine, not on the JVM.  We should be calling it JVM memory everywhere and the numbers should be 8G for public nets and 4G for private ones as a minimum

grantnoble: Thanks for your help on this @EdJoJob. Have updated the JVM numbers and removed the RAM section.

docs/HowTo/Get-Started/System-Requirements.md
```diff
@@ -25,10 +25,23 @@ To determine system requirements, check CPU and disk space requirements using
     CPU requirements are highest when syncing to the network and typically reduce after the node is
```
NicolasMassart: Should it be "higher"

Issue #291, opened 2020-03-31
TRACE is missing from this API list


NicolasMassart:
TRACE is missing from the list in https://besu.hyperledger.org/en/latest/Reference/CLI/CLI-Syntax/#rpc-http-api

PR #292, opened 2020-04-01
Added TRACE option for rpc-http-api

grantnoble:
Also sorted the API options into alphabetical order.

Fixes issue #291


grantnoble:
Linkchecker failed because of a dead link at Oracle relating to their JDKs. I checked the links to all their JDKs seem to be failing. Perhaps we wait a bit to see if they fix them?

PR #283, opened 2020-03-30
Difference between full node and archive node

grantnoble:
Include a section (in Pruning.md) about the difference between the two nodes and how to run them. Misc changes to the various pruning options and sync mode options.

Fixes #262

PR #293, opened 2020-04-01
Full and archive nodes

grantnoble:
This combines the changes made in PR #283

### DIFF
--- a/docs/Concepts/Pruning.md
+++ b/docs/Concepts/Pruning.md
@@ -4,11 +4,12 @@ description: Full and archive node types
 
 # Full and archive node types
 
-Besu supports two node types: _full nodes_ and _archive nodes_.
+Besu supports two node types, commonly referred to as _full nodes_ and _archive nodes_.
 
-Full nodes have a complete history of every block and every transaction, all fully validated, and
-can serve the network with any data request. With a full node you can check balances, sign and send
-transactions, and look at current Dapp data.
+Full nodes have just the current state of the blockchain so cannot serve the network with all data
+requests (for example, the balance of an account as at an old block). Full nodes can guarantee the
+latest state for the blockchain (and some older states, but not all). You can check current
+balances, sign and send transactions, and look at current Dapp data.
 
 Archive nodes have all of this and they also store the intermediary state of every account and
 contract for every block since the genesis block. With an archive node you can do everything you
@@ -17,17 +18,26 @@ do with a full node, as well as access historical state data.
 Archive nodes require significantly more disk space (approximately 3TB) than full nodes
 (approximately 750GB).
 
-To run an archive node, disable both fast synchronization and pruning.
+To run an archive node, enable full synchronization using
+[`--sync-mode=FULL`](../Reference/CLI/CLI-Syntax.md#sync-mode), which by default also disables
+pruning ([`--pruning-enabled=false`](../Reference/CLI/CLI-Syntax.md#pruning-enabled)).
 
-If you enable fast synchronization
-(using [`--sync-mode=FAST`](../Reference/CLI/CLI-Syntax.md#sync-mode)), pruning
-(using [`--pruning-enabled=true`](../Reference/CLI/CLI-Syntax.md#pruning-enabled)), or both, the
-node does not store all intermediary state and runs as a full node.
+To run a full node, enable fast synchronization using
+[`--sync-mode=FAST`](../Reference/CLI/CLI-Syntax.md#sync-mode), which by default also enables
+pruning ([`--pruning-enabled=true`](../Reference/CLI/CLI-Syntax.md#pruning-enabled)).
 
-## Fast sychronization
+## Fast synchronization
 
-Instead of starting from the genesis block and reprocessing all transactions, fast synchronization
-downloads the transaction receipts along with the blocks.
+Instead of a full synchronization ([`--sync-mode=FULL`](../Reference/CLI/CLI-Syntax.md#sync-mode)),
+which starts from the genesis block and reprocesses all transactions, fast synchronization
+([`--sync-mode=FAST`](../Reference/CLI/CLI-Syntax.md#sync-mode)) downloads the transaction receipts
+along with the blocks.
+
+!!! note
+
+    Fast synchronization enables pruning by default, but you can explicitly disable pruning
+    using [`--pruning-enabled=false`](../Reference/CLI/CLI-Syntax.md#pruning-enabled). Full
+    synchronization disables pruning by default.
 
 ## Pruning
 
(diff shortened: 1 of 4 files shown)


docs/HowTo/Send-Transactions/Revert-Reason.md
```diff
@@ -49,7 +49,7 @@ the revert reason as an ABI-encoded string.
     The revert reason is not included in the transactions receipts root hash. Not including the
     revert reason in the transactions receipts root hash means the revert reason is only available
     to nodes that execute the transaction when importing the block. That is, the revert reason is
-    not available if using fast sync.
+    not available if using fast synchronization (`--sync-mode=FAST`).
```
NicolasMassart, line 52: maybe add a link to ../../Reference/CLI/CLI-Syntax.md#sync-mode ?

docs/Concepts/Pruning.md
```diff
@@ -1,25 +1,61 @@
...
-description: Pruning
+description: Full and archive node types
 ---
 
-# Pruning
+# Full and archive node types
+
+Besu supports two node types, commonly referred to as _full nodes_ and _archive nodes_.
+
+Full nodes have just the current state of the blockchain so cannot serve the network with all data
```
timbeiko, line 9: I would remove the `just` to reflect that they can have more than the current state.

docs/Concepts/Pruning.md
```diff
@@ -1,25 +1,61 @@
...
+latest state for the blockchain (and some older states, but not all). You can check current
+balances, sign and send transactions, and look at current Dapp data.
+
+Archive nodes have all of this and they also store the intermediary state of every account and
+contract for every block since the genesis block. With an archive node you can do everything you
+do with a full node, as well as access historical state data.
+
+Archive nodes require significantly more disk space (approximately 3TB) than full nodes
+(approximately 750GB).
+
```
timbeiko, line 20: Maybe one thing we should add is that it is possible to re-generate all the data for an archive node using a full node (it's long, inconvenient, etc., but the data is not *lost*).

docs/Concepts/Pruning.md
```diff
@@ -1,25 +1,61 @@
...
+Full nodes have just the current state of the blockchain so cannot serve the network with all data
+requests (for example, the balance of an account as at an old block). Full nodes can guarantee the
+latest state for the blockchain (and some older states, but not all). You can check current
+balances, sign and send transactions, and look at current Dapp data.
+
+Archive nodes have all of this and they also store the intermediary state of every account and
+contract for every block since the genesis block. With an archive node you can do everything you
+do with a full node, as well as access historical state data.
+
+Archive nodes require significantly more disk space (approximately 3TB) than full nodes
```
timbeiko, line 18: Maybe worth adding these are Besu numbers, as they differ from other client implementations, especially for full nodes. We should also precise this is for mainnet.

docs/Concepts/Pruning.md
```diff
@@ -1,25 +1,61 @@
...
+
+To run a full node, enable fast synchronization using
+[`--sync-mode=FAST`](../Reference/CLI/CLI-Syntax.md#sync-mode), which by default also enables
+pruning ([`--pruning-enabled=true`](../Reference/CLI/CLI-Syntax.md#pruning-enabled)).
+
+## Fast synchronization
+
+Instead of a full synchronization ([`--sync-mode=FULL`](../Reference/CLI/CLI-Syntax.md#sync-mode)),
+which starts from the genesis block and reprocesses all transactions, fast synchronization
+([`--sync-mode=FAST`](../Reference/CLI/CLI-Syntax.md#sync-mode)) downloads the transaction receipts
```
timbeiko, line 33: Doesn't it also verify some blocks? I thought it did verify 1/n blocks, see: https://github.com/ethereum/go-ethereum/pull/1889. I'm not sure if our implementation is the same, though. cc: @RatanRSur 

Also, it's worth highlighting that past the pivot block, fast sync switches to full sync.

docs/Concepts/Pruning.md
```diff
@@ -1,25 +1,61 @@
...
 
-    If using fast sync in v1.4.0, explicitly disable pruning using
-    [`--pruning-enabled=false`](../Reference/CLI/CLI-Syntax.md#pruning-enabled).
+    If using fast synchronization
+    ([`--sync-mode=FAST`](../Reference/CLI/CLI-Syntax.md#sync-mode)) in v1.4.0, explicitly disable
+    pruning using [`--pruning-enabled=false`](../Reference/CLI/CLI-Syntax.md#pruning-enabled).
 
-Use pruning to reduce storage required for the world state. Pruning removes state trie nodes that
-are not required.
+Pruning reduces the storage required for the world state, removing state trie nodes that are not
```
timbeiko, line 53: Doesn't it reduce the storage required by removing state trie nodes from previous states? cc: @RatanRSur

docs/Concepts/Pruning.md
RatanRSur: It verifies the chain of headers from the genesis until a block when it switches to processing blocks in their entirety (full sync as you mentioned).

I would mention that it downloads the headers but I wouldn't mention that it switches to full sync at because it's an implementation detail and isn't really something they can rely on. The guarantee we give them is that they'll have the full current state but we don't want them assuming they'll have full state from any previous block because the pivot block switches many times while downloading the world state.

docs/HowTo/Send-Transactions/Revert-Reason.md
grantnoble: Done.

docs/Concepts/Pruning.md
grantnoble: Done.

docs/Concepts/Pruning.md
grantnoble: What is the process for this? If we say you can do it, we should document how.

docs/Concepts/Pruning.md
grantnoble: Done.

docs/Concepts/Pruning.md
grantnoble: Done.

docs/Concepts/Pruning.md
grantnoble: Excellent!

docs/Concepts/Pruning.md
timbeiko: It's not trivial, not something that most people will want to do (easier to just sync an archive node), but it just speaks to the security properties of a full node. Thinking about it again, I'm not sure how relevant it is here 😅

grantnoble: OK. We'll leave it as is. Thanks!

PR #323, opened 2020-04-16
Note about Prometheus and host-whitelist

### LINKED ISSUE #321: host-whitelist issue for pull metrics via prometheus
@joshuafernandes commented on [Tue Mar 31 2020](https://github.com/hyperledger/besu/issues/615)

  

### Description
As a Devops engineer, I would like to view metrics on a dashboard so I can make sure things are running smoothly

### Acceptance Criteria
* Metrics can be pulled when I specify values other than ['*'] for `host-whitelist`

### Steps to Reproduce (Bug)
1. Checkout https://github.com/joshuafernandes/besu-sample-networks/tree/test-nct 
2. Branch `test-nct`
3. Run `./run.sh -c ibft2`

**Expected behavior:** 
When specifying  host-whitelist=['localhost', '127.0.0.1']  here https://github.com/joshuafernandes/besu-sample-networks/blob/test-nct/config/besu/config.toml#L10 the nodes connect and blocks are produced but I dont see any metrics in prometheus or grafana. Updated to add 'prometheus' to that list and expected it to work but still got nothing. Even created a network with fixed IPs for each container and add that IP to the list  host-whitelist=["prometheus","172.24.2.16","localhost","127.0.0.1"] and got nothing going to prometheus. The only thing that seems to work is [*].  Note that at the nodes and chain progresses fine, this is just applicable to metrics being pulled from the prometheus container. Is this bad config I've got or a bug?

**Actual behavior:** 
Am able to get metrics only with a host-whitelist of ['*']

**Frequency:** 
100%

### Versions 
* Software version: latest as at 31/3/2020
* Java version: 11
* OS Name & Version: ubuntu 18.04 host - running docker containers via compose
* Virtual Machine software & version: n/a
* Docker Version: 19.03.7, build 7141c199a2
* Cloud VM, type, size: n/a

### Additional Information
cc: @EdJoJob @jakehaugen 

---

@magooster commented on [Mon Apr 06 2020](https://github.com/hyperledger/besu/issues/615#issuecomment-609596201)

The host whitelist is looking at the Host header of the request, for prometheus this is the 'targets' you specify in the scrape config.

scrape_configs:
    - job_name: bootnode
      scrape_interval: 15s
      scrape_timeout: 10s
      metrics_path: /metrics
      scheme: http
      static_configs:
        - targets: [ bootnode:9545 ]

host-whitelist=["bootnode","localhost","127.0.0.1"]

Unfortunately the documentation for the host-whitelist option is not completely clear. If you expose RPC services externally then you need to add the appropriate hostname or ip to the host-whitelist. By default a Besu node only allows JSON-RPC requests from localhost/127.0.0.1.

---

@shemnon commented on [Tue Apr 07 2020](https://github.com/hyperledger/besu/issues/615#issuecomment-610079963)

I'm 80% certain that we are getting a surprising host name (or no host name).  The #663 patch will log to trace every time we deny a host.  So we can crank the logging up to trace and then grep the log (or do fancy log4j stuff) and see if we get strange host names.  If there is no `host:` header then prometheus is not sending a host for us to whitelist off of.  In that case we need to consider other options.

---

@ajsutton commented on [Tue Apr 07 2020](https://github.com/hyperledger/besu/issues/615#issuecomment-610083311)

+1 to finding out exactly what's going on first and we probably should validate this with real experts, but...

I believe it would be safe to allow requests with no `host` header.  The key reason for the domain whitelisting is to prevent DNS redirection attacks which allow any website to send a request to `mycustomdomain.com` and have `mycustomdomain.com` resolve to `127.0.0.1` and thus allow the request to be sent to the node.

Browsers always include a `host` header so requests without it didn't come from such an uncontrolled source.

If we're being paranoid (and we probably should be) then supporting a `<none>` option in the whitelist might be a suitable option as well.

---

@magooster commented on [Tue Apr 07 2020](https://github.com/hyperledger/besu/issues/615#issuecomment-610095281)

Prometheus scraper definitely sets a Host header amongst others including User-Agent, X-Prometheus-Scrape-Timeout-Seconds, Accept-Encoding & Accept. In my testing Host header defaults to the URL used in the scrape-config.

If you expose an RPC service externally you'll need to add the hostname and/or ip of the external interface to the host-whitelist.

Rather than logging why not add the hostname to the 403 response (I don't think this discloses anything significant - especially if I already can find the endpoint)

Host <hostname> not authorized

---

@shemnon commented on [Wed Apr 08 2020](https://github.com/hyperledger/besu/issues/615#issuecomment-610772519)

Will Prometheus surface what it gets in the 403 bodies?  I considered that but since it was prom only I was concerned the output would be dropped.

---

@magooster commented on [Wed Apr 08 2020](https://github.com/hyperledger/besu/issues/615#issuecomment-610805540)

That's a fair point - you only get the scraper logs if you set prometheus log level to debug and then you only see a 403. You should in theory be able see each endpoint status from the prometheus UI. (prometheus fqdn/ip:9090/targets), but for the sample networks those are internal endpoints.

Two additional points when it comes to sample-networks and monitoring

For the sample networks host-whitelist would need to specified for each container.

The prometheus config is not targeted to each network, so there's a bunch of  redundant scrapers.

---

@joshuafernandes commented on [Wed Apr 15 2020](https://github.com/hyperledger/besu/issues/615#issuecomment-613803122)

So as per @magooster 's comments the scraper seem's to use the host its scraping as the host header and setting the host-whitelist to the node's name resolves this behaviour. That being said the scraper doing that for the host header is a bit odd in the pull context...  

Prometheus' logs have nothing in there for failed reqs

---

@MadelineMurray commented on [Thu Apr 16 2020](https://github.com/hyperledger/besu/issues/615#issuecomment-614339652)

As discussed with @joshuafernandes this behaviour needs to be doc'd so changing this to a docs issue.

grantnoble:
Included a note that if using Prometheus to pull metrics from a node, you must specify all the other nodes you want to pull metrics from in the whitelist.
Fixes #321

### DIFF
--- a/docs/Reference/CLI/CLI-Syntax.md
+++ b/docs/Reference/CLI/CLI-Syntax.md
@@ -341,6 +341,11 @@ A comma-separated list of hostnames to allow
 [access to the JSON-RPC API](../../HowTo/Interact/APIs/Using-JSON-RPC-API.md#host-whitelist). By
 default, Besu accepts access from `localhost` and `127.0.0.1`.
 
+!!!note
+
+    If using [Prometheus](https://prometheus.io/) to pull metrics from a node, you must specify all
+    the other nodes you want to pull metrics from in the whitelist of hostnames.
+
 !!!tip
 
     To allow all hostnames, use `"*"`. We don't recommend allowing all hostnames for production


grantnoble:
Not entirely sure if this is worthy of just a note, or whether it should be a caution or a warning?

PR #324, opened 2020-04-16
Clarify the rpc-*-authentication-jwt-public-key-file options

grantnoble:
Clarify that the `http` and `ws` versions of these options specify the JWT provider's public key file. Also, clarify these points in the authentication steps, and that generation of a private/public key pair is for demonstration purposes only. Also, tidy up references to JSON Web Tokens (JWTs) instead of JWT tokens. And finally a few minor fixes.

Fixes #314 

Signed-off-by: grantnoble <grant.noble@consensys.net>

PR #327, opened 2020-04-17
Change JWT token to JWT

grantnoble:
As in PR #324, references to "JWT token" changed to "JWT". "JSON Web Token (JWT)" in the first instance, then "JWT" from then on.

### DIFF
--- a/docs/Tutorials/Privacy/Configuring-Multi-Tenancy.md
+++ b/docs/Tutorials/Privacy/Configuring-Multi-Tenancy.md
@@ -34,8 +34,9 @@ IBFT-Network/
 
 ## 1. Generate a private and public key pair
 
-In the `Node-1` directory [create the `.pem` formatted key pair]. The key pair belongs to the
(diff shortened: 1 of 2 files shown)


docs/Tutorials/Privacy/Configuring-Multi-Tenancy.md
```diff
@@ -35,13 +35,12 @@ IBFT-Network/
 ## 1. Generate a private and public key pair
 
 In the `Node-1` directory [create the `.pem` formatted key pair]. The key pair belongs to the
```
MadelineMurray, line 37: This wasn't introduced by this PR but the link isn't working properly here - I can see it at the bottom of the page but it doesn't publish as a link.  Could you fix that as part of this PR.

grantnoble: It doesn't like the back ticks used to format ".pem". I resturctured the link, and put the ".pem" as part of the next sentence, outside of the link.

PR #373, opened 2020-05-18
Added private filters concept content

MadelineMurray:
Added private filters concept content


docs/HowTo/Interact/Filters/Accessing-Logs-Using-JSON-RPC.md
```diff
@@ -150,10 +154,24 @@ When a filter is no longer required, use
...
+
+Filters for private contracts are created, accessed, and uninstalled using: 
+
+* [`priv_getFilterChanges`](../../../Reference/API-Methods.md#priv_getfilterchanges)
+* [`priv_getFilterLogs`](../../../Reference/API-Methods.md#priv_getfilterlogs)
+* [`priv_getLogs`](../../../Reference/API-Methods.md#priv_getlogs)
+* [`priv_newFilter`](../../../Reference/API-Methods.md#priv_newfilter)
+* [`priv_uninstallFilter`](../../../Reference/API-Methods.md#priv_uninstallfilter).  
+  
+The [privacy group ID](../../../Concepts/Privacy/Privacy-Overview.md) must be specified as a parameter
```
lucassaldanha, line 167: Should we clarify that the privacyGroupId is the parameter index 0 in the request or will we have a request example somewhere in the docs?

MadelineMurray: If you click through to the reference material you'll see examples but it's probably a good idea to include one here as well so I'll add that.

PR #382, opened 2020-05-27
change wording since pruning is not currently the default, even with fast sync

RatanRSur:


RatanRSur:
I guess I'm not a maintainer here? xD

RatanRSur:
@NicolasMassart Is this waiting on something else?

PR #390, opened 2020-06-08
Added tx-pool-price-bump

MadelineMurray:
Added tx-pool-price-bump


docs/Reference/CLI/CLI-Syntax.md
```diff
@@ -2260,6 +2260,26 @@ tx-pool-hashes-max-size="2000"
...
+```bash tab="Command Line"
+--tx-pool-price-bump=25
+```
+
+```bash tab="Environment Variable"
+BESU_TX_POOL_PRICE_BUMP=25
+```
+
+```bash tab="Configuration File"
+tx-pool-price-bump="25"
```
bgravenorst, line 2278: Do we need quotes for integer values?

MadelineMurray: @abdelhamidbakhta - I tried this with and without quotes and Besu started both times but I assume the value was ignored in one case.  Can you confirm if format for the config file?

abdelhamidbakhta: You can with and without quotes. Besu will start and the value will be used. The value is never ignored. If it is an integer between 0 and 100 it works, with and without quotes. If the value is not and integer or not within the range Besu will fail with an explicit error.

PR #402, opened 2020-06-14
Removed cautions stating that pruning needs to be explicitly disabled

MadelineMurray:
Removed cautions stating that pruning needs to be explicitly disabled (it's disabled by default)


NicolasMassart:
Youtube link failed, it's probably a temporary issue, but I take this opportunity to say that we should use the short youtube URL provided by the share button on their site taht looks like `https://youtu.be/7BuutRe0I28` instead of `https://www.youtube.com/watch?v=7BuutRe0I28&feature=youtu.be`

PR #404, opened 2020-06-15
Update examples requesttimeoutseconds and fork

### LINKED ISSUE #401: Sample genesis file for ibft2
## Describe the bug

I see two issues with the sample genesis file, its not using the latest protocol hardfork -> 'constantinoplefixblock' line, and the 'requesttimeoutseconds' settings does not follow guidance. 

The former means the chain doesn't use the latest set of precompiles and gas costs, (no chainId support in contracts), and the latter means if a validator goes offline you see a 10s gap between some blocks. (The recommendation is to set this to twice the block period seconds). 

Check the issue type in the following list (insert X instead of space between [ ]): -->

- [ ] Missing content
- [ x] Outdated content
- [ ] Wrong content
- [ x] Confusing or misleading content
- [ ] Other

## The broken page

https://besu.hyperledger.org/en/stable/Tutorials/Private-Network/Create-IBFT-Network/

## Change suggestion

Use the latest supported hard fork, 'muirglacierblock' and set the requesttimeoutseconds to 4

## More context

The same genesis is used with the besu sample networks.

'constantinoplefixblock' is used across all sample genesis files...

NicolasMassart:
make requesttimeoutseconds double of blockperiodseconds
use latest muirglacierblock fork features

fixes #401


docs/HowTo/Configure/Consensus-Protocols/IBFT.md
```diff
@@ -122,8 +122,8 @@ blocks every `blockperiodseconds`.
 
     An internal PegaSys IBFT 2.0 testnet has four geographically dispersed validators in Sweden,
     Sydney, and two in North Virginia. With a `blockperiodseconds`of 5 and a
-    `requesttimeoutseconds` of 10, the testnet consistently creates blocks with a five second
-    blocktime.
+    `requesttimeoutseconds` of 10 (`blockperiodseconds` x 2), the testnet consistently creates
```
rain-on, line 125: 1. The network is not longer running, so may need to change
2. Where is our guidance of reqeusttimeout=2xblocktime? Its not a _bad_ suggestion,  but its probably better to suggest that the numbers should be tuned to meet your network performance, and required block propogation rate.

NicolasMassart: I updated with suggestion to requesttimeout = block time + (some network communication x number of nodes)

docs/HowTo/Configure/Consensus-Protocols/IBFT.md
```diff
@@ -113,17 +115,33 @@ Usually, the protocol adds the proposed block before reaching `requesttimeoutsec
...
+- Set `blockperiodseconds` to your desired block time, the waiting duration between blocks.
+- Set `requesttimeoutseconds` depending on your network and the number of IBFT2 validator nodes.
+    This value requires tuning for your specific environment, but you can calculate a decent initial value
+    to obtain blocks every `blockperiodseconds` with the following formula:
 
-    An internal PegaSys IBFT 2.0 testnet has four geographically dispersed validators in Sweden,
-    Sydney, and two in North Virginia. With a `blockperiodseconds`of 5 and a
-    `requesttimeoutseconds` of 10 (`blockperiodseconds` x 2), the testnet consistently creates
-    blocks with a five second blocktime.
+    `requesttimeoutseconds = blockperiodseconds + (validators count * network delay)`
```
rain-on, line 128: I'd personally avoid making any hard-fast statements of how to calculate it - not sure on our position regarding making deployment style.
Can we say  "reduce this number until you starting having "round changes" then increase until round change is avoided?

docs/HowTo/Configure/Consensus-Protocols/IBFT.md
```diff
@@ -113,17 +121,22 @@ Usually, the protocol adds the proposed block before reaching `requesttimeoutsec
...
+
+#### Tuning block timeout 
+
+To tune the block timeout for your network deployment: 
 
-    An internal PegaSys IBFT 2.0 testnet has four geographically dispersed validators in Sweden,
-    Sydney, and two in North Virginia. With a `blockperiodseconds`of 5 and a
-    `requesttimeoutseconds` of 10, the testnet consistently creates blocks with a five second
-    blocktime.
+1. Start by setting `blockperiodseconds` to your desired block time and `requesttimeoutseconds` to two times
```
bgravenorst, line 136: Perhaps rephrase to, "Set `blockperiodseconds` to..."

docs/HowTo/Configure/Consensus-Protocols/IBFT.md
MadelineMurray: Updated per Trent's feedback

docs/HowTo/Configure/Consensus-Protocols/IBFT.md
MadelineMurray: Done

PR #454, opened 2020-07-26
Add new txpool API.

bgravenorst:
Signed-off-by: Byron Gravenorst <byron.gravenorst@consensys.net>

## Describe the change

Added the txpool_besuPendingTransactions API.


docs/Reference/API-Methods.md
```diff
@@ -4390,6 +4390,109 @@ None
...
+### txpool_besuPendingTransactions
+
+Lists pending transactions that match the supplied filter conditions.
+
+
+#### Parameters
+
+* `QUANTITY` - Integer representing the maximum number of results to return.
+* Object containing the filter conditions. Filter pending transactions based on the following fields
+    and supported operators:
```
shemnon, line 4402: I think some fo the filter conditions need to be explicitly spelled out.
```suggestion
* Object containing the filter conditions. 
  Each field in the object corresponds to a field, and the value is a single field object where the field name is the operator and the field value is the value for the operator.  Multiple transactions fields can be provided, and all filters must be satisfied for a transaction to be returned.  Each field can have only one operator (i.e. you cannot ask for transactions whose gas price is between 8 and 9 GWei, by using both the `gt` and `lt` operator).
```

bgravenorst: @shemnon can we supply the same field twice? For example:

```
             "gas":{
                "lt":"0x5209"
             },
             "gas":{
                "gt":"0x4000"
             }
```

bgravenorst: Actually, ignore^^, I just tested it with the following and it worked fine.

```
{
   "jsonrpc":"2.0",
   "method":"txpool_besuPendingTransactions",
   "params":[
      2,
      {
         "gas":{
            "lt":"0x5209"
         },
         "gas":{
            "gt":"0x4209"
         },
         "nonce":{
            "gt":"0x1"
         }
      }
   ],
   "id":1
}
```

bgravenorst: I've reworded you suggestion slightly.

docs/Reference/API-Methods.md
```diff
@@ -4390,6 +4390,116 @@ None
...
+Lists pending transactions that match the supplied filter conditions.
+
+
+#### Parameters
+
+* `QUANTITY` - Integer representing the maximum number of results to return.
+* Object of fields used to create the filter condition.
+
+Each field in the object corresponds to a field name containing an operator, and a value for the
+operator. The same field name can be specified multiple times, but can only contain one operator.
```
MadelineMurray, line 4404: I think this makes sense but I think it would be good to include an example where the same field name is specified multiple times.  ie, From reading this I assume you can have a request with params such as: 

 "params":[
          2,
          {
             "from":{
                "eq":"0xfe3b557e8fb62b89f4916b721be55ceb828dbd73",
                "eq":"0x627306090abaB3A6e1400e9345bC60c78a8BEf57"
             },
             "gas":{
                "lt":"0x5209"
             },
             "nonce":{
                "gt":"0x1"
             }
          }
       ],

docs/Reference/API-Methods.md
shemnon: That shouldn't work.  I would expect that only one of the "gas" values will have an effect, since it goes into a hashmap.

docs/Reference/API-Methods.md
shemnon: Also, this won't work.  We will only get one of the EQ fields.  if we tried GT and LT in the same operator it will fail because we explicitly check for it.

docs/Reference/API-Methods.md
bgravenorst: @shemnon, should we have some sort of check for this? i.e Duplicate field names.

shemnon: It's actually bad JSON.  https://tools.ietf.org/html/rfc8259#section-4 

>The names within an object SHOULD be unique.

The particular details of our Library obscures this from us before we can even get a chance to detect it.  It puts the fields in a hash map.

PR #467, opened 2020-08-06
fixed some links and clarified priv_getFilterChanges description

macfarla:
Signed-off-by: Sally MacFarlane <sally.macfarlane@consensys.net>

## Describe the change

Fixed several links, and also clarified the description for priv_getFilterChanges

### For content changes

- [ x] Doc content
- [ ] Doc pages organisation

### DIFF
--- a/docs/Reference/API-Methods.md
+++ b/docs/Reference/API-Methods.md
@@ -5145,12 +5145,6 @@ Distributes a signed, RLP encoded
 Returns the private transaction count for the specified account and
 [group of sender and recipients].
 
-!!! important
-
-    If sending more than one transaction to be mined in the same block (that is, you are not
-    waiting for the transaction receipt), you must calculate the private transaction nonce outside
-    Besu.
-
 #### Parameters
 
 `data` - Account address.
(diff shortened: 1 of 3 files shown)


docs/Reference/API-Methods.md
```diff
@@ -5145,12 +5145,6 @@ Distributes a signed, RLP encoded
 Returns the private transaction count for the specified account and
 [group of sender and recipients].
 
-!!! important
-
-    If sending more than one transaction to be mined in the same block (that is, you are not
-    waiting for the transaction receipt), you must calculate the private transaction nonce outside
-    Besu.
-
```
macfarla, line 5153: I removed this because I was thinking it's not relevant to priv_getEeaTransactionCount (nonce isn't an input to this function) but is it meant to mean "use this method to calculate the nonce if you need to"?

MadelineMurray: It actually means the opposite - if you're sending more than one transaction per block, this function won't return the correct nonce for you (ie, don't use it).   We could update it to: 

If sending more than one transaction to be mined in the same block (that is, you are not
waiting for the transaction receipt), you must calculate the private transaction nonce outside
Besu instead of using `priv_getEeaTransactionCount`.

macfarla: done

PR #481, opened 2020-08-10
Add note to eth_getLogs

timbeiko:
Adds a note to the `eth_getLogs` API to recommend not using it for large block ranges. 

Once this is in, I believe we can remove [this](https://github.com/hyperledger/besu/blob/master/KNOWN_ISSUES.md#scope-of-logs-query-causing-besu-to-hang) from the Besu known issues.


timbeiko:
Good catch, thanks @bgravenorst 😄

PR #500, opened 2020-08-28
Updated group management contract ownership doc

MadelineMurray:
Updated group management contract ownership doc


docs/Concepts/Privacy/Onchain-PrivacyGroups.md
```diff
@@ -30,8 +30,9 @@ create a privacy group, the contract bytecode is part of the genesis state of th
     All members of an onchain privacy group must be using the same version of Hyperledger Besu. If
     using different versions, the private state within the privacy group may become inconsistent.
 
-The provided group management contracts do not enforce any permissioning on group members. That is,
-any member of a group can make transactions in the group, and add or remove group members.
+In the provided group management contract the owner of the privacy group management contract is the signer
+of the private transaction that created the privacy group. In the default group management contract
+implementation, only the owner can add and remove participants, and upgrade the management contract.
```
macfarla, line 35: is "provided group management contract" == "default group management contract"

macfarla: How about:
In the default implementation of the group management contract, the signer of the private transaction that creates the privacy group is also the owner of the group. Only the owner can add and remove participants, and upgrade the management contract.

macfarla (commented):
I tried to hemingway it

docs/Concepts/Privacy/Onchain-PrivacyGroups.md
MadelineMurray: Thanks - that's much better! Have updated.

macfarla (approved):
LGTM

docs/Concepts/Privacy/Onchain-PrivacyGroups.md
```diff
@@ -30,8 +30,9 @@ create a privacy group, the contract bytecode is part of the genesis state of th
     All members of an onchain privacy group must be using the same version of Hyperledger Besu. If
     using different versions, the private state within the privacy group may become inconsistent.
 
-The provided group management contracts do not enforce any permissioning on group members. That is,
-any member of a group can make transactions in the group, and add or remove group members.
+In the default implementation of the group management contract, the signer of the private transaction
+that creates the privacy group is also the owner of the group. Only the owner can add and remove participants,
```
pinges, line 34: Do we need to be very specific here and say something like: " ... only private transactions signed by the same private key as the private transaction that created the privacy group can add and ... "
I do understand what you are saying, but I think we should be specific about the signing key ?

macfarla: maybe a "note" saying "what this means at a technical level is..."

MadelineMurray: Added an extra paragraph: 

```
The owner is identified by the signing key.  Transactions to add and remove participants, or upgrade 
the management contract, must be signed by the same key that signed the group creation transaction.
```

Does that work?

macfarla: 👍

NicolasMassart (approved):
Looks good to me.

PR #516, opened 2020-09-15
Add DNS support

bgravenorst:
Signed-off-by: Byron Gravenorst byron.gravenorst@consensys.net

## Describe the change
Added documentation about using DNS in private permissioned networks.


docs/Reference/CLI/CLI-Syntax.md
```diff
@@ -3029,6 +3029,79 @@ The price bump percentage to replace an existing transaction. The default is 10.
...
+
+    ```bash
+    Xdns-enabled=true
+    ```
+
+Enables DNS support. The default is `false`.
+
+!!! important
+
+    Use domain names in private networks only because public networks require using IP addresses.
```
joshuafernandes, line 3062: Suggest updating to: `Use DNS with trusted DNS provider in private networks because of limitations where IP address can change eg Kubernetes pods`

docs/Reference/CLI/CLI-Syntax.md
```diff
@@ -3029,6 +3029,79 @@ The price bump percentage to replace an existing transaction. The default is 10.
...
+
+    This is an early access feature. Some functionality may be updated before the feature is fully
+    released.
+
+You can use DNS when using Kubernetes in a private permissioned network to ensure you can reconnect
```
joshuafernandes, line 3067: Combine this line with the one above in 3062?

joshuafernandes (approved):
LGTM - suggested one change

docs/Reference/CLI/CLI-Syntax.md
bgravenorst: Done

docs/Reference/CLI/CLI-Syntax.md
bgravenorst: Done

docs/Concepts/Node-Keys.md
```diff
@@ -42,6 +42,13 @@ You can export the node address, either to standard output or to a specified fil
 The enode URL identifies a node. For example, the `--bootnodes` option and the
 `perm_addNodesToAllowlist` method specify nodes by enode URL.
 
+!!! tip
+
+    If deploying Besu using Kubernetes in private permissioned network, use the
+    [`--Xdns-enabled`](../Reference/CLI/CLI-Syntax.md#xdns-enabled) option to use domain names
+    instead of IP addresses. This ensures that Besu can connect to a container even if the IP
```
matkt, line 49: I would say that this option alone does not do that. With this option, Besu will query the DNS when it starts up and if the container IP changes afterwards it will not detect it. You must add the second option `Xdns-update-enabled` so that Besu query the DNS each time and detect a change of IP

matkt (changes requested):

docs/Concepts/Node-Keys.md
bgravenorst: Done.

PR #518, opened 2020-09-16
Updated net_version

MadelineMurray:
Updated net_version


shemnon (commented):
I think we need some discussion about the fact that for `classic` and `mordor` that the networkId is different from the ChainId, and maybe call out the ChainId values of those two networks.  Otherwise I expect someone will configure their metamask off of the networkId despite the link to the page on networkId vs chainId.

docs/Reference/API-Methods.md
```diff
@@ -501,26 +501,26 @@ is a [Keccak-256](https://keccak.team/keccak.html) hash, not the standardized SH
...
+| Network ID | Chain | Network | Description
+|------------|-------|---------|-------------------------------|
+| `1`        | ETH   | MainNet | Main Ethereum network         |
+| `3`        | ETH   | Ropsten | PoW test network              |
+| `4`        | ETH   | Rinkeby | PoA test network using Clique |
+| `5`        | ETH   | Goerli  | PoA test network using Clique |
+| `2018`     | ETH   | Dev     | PoW development network       |
+| `1`        | ETC   | Classic | Main Ethereum Classic network |
+| `6`        | ETC   | Kotti   | PoA test network using Clique |
+| `7`        | ETC   | Mordor  | PoW test network              |
```
shemnon, line 524:
```suggestion

!!! note

    For almost all networks Network ID and Chain ID are the same.  
    The only networks where it differs currently are Classic with a Chain ID of `61` and Mordor with a Chain ID of `63`.

```

MadelineMurray: Added

NicolasMassart (approved):
LGTM.
Link checker tests will pass once #521 merged in master and this PR updated. It may require a few markdown fixes because of the new rules, but probably no more than whitespaces fixes.

PR #584, opened 2020-12-20
Updated section on confirming if network is working

MadelineMurray:
Updated section on confirming if network is working


bgravenorst (approved):
Looks good to me

PR #586, opened 2020-12-29
Updated eth_gasPrice and added api-gas* cli options

MadelineMurray:
Updated eth_gasPrice and added api-gas* cli options

### DIFF
--- a/docs/Reference/API-Methods.md
+++ b/docs/Reference/API-Methods.md
@@ -1042,7 +1042,7 @@ None
 
 ### `eth_gasPrice`
 
-Returns a percentile gas unit price for the most recent blocks, in Wei. By default, 
+Returns a percentile gas unit price for the most recent blocks, in Wei. By default,
 the last 100 blocks are examined and the 50th percentile gas unit price (that is, the median value)
 is returned.
 
@@ -1053,7 +1053,7 @@ and [`--api-gas-price-max`](CLI/CLI-Syntax.md#api-gas-price-max). By default, 10
 
 Use the [`--api-gas-price-blocks`](CLI/CLI-Syntax.md#api-gas-price-blocks), [`--api-gas-price-percentile`](CLI/CLI-Syntax.md#api-gas-price-percentile)
 , and [`--api-gas-price-max`](CLI/CLI-Syntax.md#api-gas-price-max) command line
-options to configure the `eth_gasPrice` default values. 
+options to configure the `eth_gasPrice` default values.
 
 #### Parameters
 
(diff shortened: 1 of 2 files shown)


docs/Reference/API-Methods.md
```diff
@@ -1042,17 +1042,26 @@ None
...
-Returns the current gas unit price, in wei. It's the hexadecimal equivalent of the price specified
-for the [`--min-gas-price`](CLI/CLI-Syntax.md#min-gas-price) command line option when the node
-started, or the default minimum gas price.
+Returns a percentile gas unit price for the most recent blocks, in Wei. By default, 
+the last 100 blocks are examined and the 50th percentile gas unit price (that is, the median value)
+is returned.
+
+If there are no blocks, the value for [`--min-gas-price`](CLI/CLI-Syntax.md#min-gas-price) is returned.
+The value returned is restricted to values between [`--min-gas-price`](CLI/CLI-Syntax.md#min-gas-price)
+and [`--api-gas-price-max`](CLI/CLI-Syntax.md#api-gas-price-max). By default, 1000 Wei and
```
MadelineMurray, line 1051: @shemnon - I've included the default min-gas-price here (1000) not the 100 from your [original PR](https://github.com/hyperledger/besu/pull/1563). Flagging in case I've got something wrong.

NicolasMassart: I confirm, this may have been a typo. Min gas price default is `Wei.of(1000)`

docs/Reference/CLI/CLI-Syntax.md
```diff
@@ -36,6 +36,98 @@ To start a Besu node run:
...
+=== "Syntax"
+
+    ```bash
+    --api-gas-price-blocks=<INTEGER>
+    ```
+
+=== "Example"
+
+    ```bash
+    --api-gas-price-blocks=1000
```
NicolasMassart, line 50: Are we sure this `1000` blocks example is not going to create any issue? It's a lot. The more blocks the more variation of gas price you can get. I tend to think that if user has to change the value to override the default, it may be with a lower value instead of a higher one to be closer to the latest gas price. A higher one would only provide a value closer to the long term median and I don't think there's much interest for that as the goal is to be able to accurately evaluate the gas price to send a tx. Smoothing the latests changes a bit is important, but having a too wide range will provide more inaccurate values.

docs/Reference/API-Methods.md
shemnon: Yea, typo.  1000 is a fine minimum.  Mostly a flag that there have been no TXes.

shemnon (approved):

docs/Reference/CLI/CLI-Syntax.md
MadelineMurray: Good point.  Updated to 50.

PR #590, opened 2020-12-30
Added --compatibility-eth64-forkid-enabled

MadelineMurray:
Added --compatibility-eth64-forkid-enabled

### DIFF
--- a/docs/Reference/CLI/CLI-Syntax.md
+++ b/docs/Reference/CLI/CLI-Syntax.md
@@ -135,6 +135,46 @@ When connecting to MainNet or public testnets, the default is a predefined list 
 In private networks defined using [`--genesis-file`](#genesis-file) or when using
 [`--network=dev`](#network), the default is an empty list of bootnodes.
 
+### `compatibility-eth64-forkid-enabled`
+
+=== "Syntax"
+
+    ```bash
+    --compatibility-eth64-forkid-enabled[=<true|false>]
+    ```
+
+=== "Command Line"
+
+    ```bash
+    --compatibility-eth64-forkid-enabled=true
+    ```
+
+=== "Environment Variable"
+
+    ```bash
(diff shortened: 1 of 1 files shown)


docs/Reference/CLI/CLI-Syntax.md
```diff
@@ -135,6 +135,46 @@ When connecting to MainNet or public testnets, the default is a predefined list
...
+
+Enables the legacy Eth/64 fork ID. For any networks with nodes using Besu v1.4 or earlier and nodes
+using Besu v20.10.1 or later, either:
+
+* All nodes must be upgraded to v20.10.1 or later.
+* All nodes using v20.10.1 or later must have `--compatibility-eth64-forkid-enabled` set to `true`.
+
+The default is `false`.
+
+!!! caution
```
NicolasMassart, line 172: Just a thought but maybe we have to decide if we make all these "warning", "attention" and "caution" consistent. We have 3 terms for something I think is mostly the same.

MadelineMurray: Agree.  I use caution mostly I think.  Can we update the template so caution is an option but warning and attention are not?

NicolasMassart: I'll check what we can do using Vale.
Note that "caution" looks less familiar to me as a non native english reader or not the expected langage level for doc compared to warning which is commonly used in code. So that's perhaps something to consider too.

abdelhamidbakhta (approved):
Looks good to me.

docs/Reference/CLI/CLI-Syntax.md
NicolasMassart: Given what we discussed internally on the error/warning/caution levels it looks ok for me to use caution here and we will try to provide a comprehensive explanation of how to use these levels.

NicolasMassart (approved):

PR #619, opened 2021-03-02
Add information clarifying IBFT 2 network behaviour.

### LINKED ISSUE #611: Make clear what bahaviour will occur in IBFT 2 network if more than 1/3 validators stop
If more than 1/3 of the validators stopped participating, new blocks can no longer be created.  The remaining validators will continue to propose blocks and the requesttimeoutseconds doubles each time the round fails to add the proposed block. 

What will be been seen in the logs if the TRACE log level is enabled, is lots of round change log messages. 

The quickest way to restart the network producing blocks is to restart all of the validators -> this will set the requesttimeoutseconds back to the original value specified in the genesis file.

bgravenorst:
## Describe the change

Add information clarifying IBFT 2 network behaviour.

## Issue fixed
 Fixes #611


docs/HowTo/Configure/Consensus-Protocols/IBFT.md
```diff
@@ -10,7 +10,8 @@ IBFT 2.0.
 !!! warning
 
     Configure your network to ensure you never lose 1/3 or more of your validators. If more
-    than 1/3 of validators stop participating, new blocks are no longer be created.
+    than 1/3 of validators stop participating, new blocks are no longer created and
+    serious issues can occur.
```
NicolasMassart, line 14: Could it be more specific about what kind of serious issues? How serious, what can be broken?
Otherwise, perhaps "unexpected issues" would be better?

MadelineMurray: @bgravenorst - to Nico's point this could be more explicit: 

"your network will stall and may take significant time to recover once nodes are restarted."

bgravenorst: Done

NicolasMassart (approved):
Looks good to me.

PR #641, opened 2021-03-17
fix(markup): gfm defects

sambacha:
fixes issue with markup tables

besu-docs-diff


NicolasMassart (approved):
Thanks for making us aware of the table display issue in Github
We look at source code and final rendering on site but not really how Github renders the markdown as we have a lot of non github compliant markdown extensions used with mkdocs.
I kept the table fixes but made some changes so the table source code is easier to read. It doesn't change the final rendering, either in github or mkdocs.

PR #847, opened 2021-10-29
New node doc

Ezzahhh:
## Pull Request Description 
Replicating documentation from GoQuorum with relevant modifications for Besu.

### DIFF
--- a/docs/Tutorials/Developer-Quickstart.md
+++ b/docs/Tutorials/Developer-Quickstart.md
@@ -10,13 +10,13 @@ description: Rapidly generate local blockchain networks.
 - [Docker and Docker-compose](https://docs.docker.com/compose/install/)
 - [Node.js](https://nodejs.org/en/download/) version 12 or higher
 - On Windows:
-    - Windows Subsystem for Linux 2
-    - Docker desktop configured to use the WSL2-based engine
+  - Windows Subsystem for Linux 2
+  - Docker desktop configured to use the WSL2-based engine
 
 !!! important
-    Ensure you allow Docker up to 4G of memory or 6G if running the privacy examples.
-    Refer to the _Resources_ section in [Docker for Mac](https://docs.docker.com/docker-for-mac/) and
-    [Docker Desktop](https://docs.docker.com/docker-for-windows/) for details.
+Ensure you allow Docker up to 4G of memory or 6G if running the privacy examples.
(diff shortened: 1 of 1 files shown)


docs/Tutorials/Developer-Quickstart.md
```diff
@@ -65,3 +66,99 @@ the tutorial.
...
+
+### 5. Update files with the enode address
+
+Add the new node's enode address to the [static nodes] file and [permissions file].
+The enode uses the format `enode://pubkey@ip_address:30303`.
+If the `nodekey.pub` is `4540ea...9c1d78` and the IP address is `172.16.239.41`, then the enode
+address would be `"enode://4540ea...9c1d78@172.16.239.41:30303"`,
+which must be added to both files.
+
+As an alternative, it is also possible to call the Besu API method [`perm_addNodesToAllowlist`](https://besu.hyperledger.org/en/latest/Reference/API-Methods/#perm_addnodestoallowlist) to add the new node to the [allow list].
```
joshuafernandes, line 149: Suggest reword to `For the permissioned allow list, you can alternatively call the API method ..... to add the new node`

docs/Tutorials/Developer-Quickstart.md
```diff
@@ -65,3 +66,99 @@ the tutorial.
...
+
+### 6. Start the network
+
+Once complete, start the network up with `./run.sh`.
+On a live network the process is the same when using local permissions with the `permissions_config.toml` file.
+You don't need to restart the network and subsequent changes to the files are picked up by the servers.
```
joshuafernandes, line 155: Might pay to remove this line about changes being picked, not 100% sure Besu does this. And add that on the live network you have to add the node to the allowlist but you don't need to restart the node and can make the API call. The reason we add to the permissions.json is that in the event the node is restarted the API call's entry is lost

joshuafernandes (approved):
Looks good, @Ezzahhh, couple of minor suggestions

docs/Tutorials/Developer-Quickstart.md
Ezzahhh: I just removed those two lines and placed them into a note block under "5. Update files with the enode address". Let me know if you think it should be reworded further.

bgravenorst (approved):
Made a few updates. LGTM!

PR #890, opened 2021-12-07
update and add content around private nonces

### LINKED ISSUE #849: add content around private nonce management
Here's some words, as a starting point for some docs content around private nonces. Happy to discuss!

One transaction: you call getTransactionCount and then use that nonce to make a transaction
Multiple transactions: you do this twice. getTransactionCount, sendTransaction, getTransactionCount, sendTransaction. If things happen in that order, yay. Except for one small thing: if your second call to getTransactionCount happens before the first transaction is processed, you can get the same nonce again. In the public world we get around this with getTransactionCount(“pending”) ie you take into account any pending transactions. Still, there is a small window in which you might get the same nonce if the calls come in too quickly. 

Options for private nonce management:
1 - Let Besu handle it (see below). You will just need to make sure that you wait “long enough” between calls. Currently this is 1.5 seconds, depending on block time. (besu PR 2969 should reduce this time)
2 - Manage the nonce yourself - see concurrentTransactions example. This is a good option if you are sending a lot of transactions that are independent of each other.
3 - Use orchestrate for nonce handling. Enterprise use.

Example: If you are using web3js-quorum generateAndSendRawTransaction, under the hood this uses priv_getTransactionCount to derive the nonce for the private transaction. (Unless you keep track of and provide the nonce yourself.) So if you call generateAndSendRawTransaction twice, this can effectively result in priv_getTransactionCount, priv_getTransactionCount, sendTransaction, sendTransaction. The two calls in a row to priv_getTransactionCount (if you call them both within “the window”) can both return the same nonce, and one of the transactions will fail (nonce mismatch error).

Public transactions have the same issue (this is a general problem with distributed systems) but the window is small, since we can use the transaction pool to take into account transactions that are pending.
For private transactions, that window is currently longer because we are creating a private transaction and then a public transaction to wrap it (PMT) - and we wait until the PMT is included in a block (that is, we don’t currently take into account pending transactions).

- [x] Missing content

alexandratran:
Signed-off-by: Alexandra Tran <alexandra.tran@consensys.net>

## Pull request checklist

Use the following list to make sure your PR fits the Besu doc quality standard.

### Before creating the pull request

Make sure that:

- [x] [all commits in this PR are signed off for the DCO](https://wiki.hyperledger.org/display/BESU/DCO).
- [x] you read the [contribution guidelines](https://wiki.hyperledger.org/display/BESU/Contributing+to+documentation).
- [x] you have [tested your changes locally](https://wiki.hyperledger.org/display/BESU/MkDocs+And+Markdown+Guide#MkDocsAndMarkdownGuide-PreviewTheDocumentation) before submitting them to the community for review.

### After creating your pull request and tests finished

Make sure that:

- [x] you fixed all the issues raised by the tests, if any.
- [x] you verified the rendering of your changes on [ReadTheDocs.org PR preview](https://wiki.hyperledger.org/display/BESU/MkDocs+And+Markdown+Guide#MkDocsAndMarkdownGuide-PreviewwithReadTheDocs)
  and updated the testing link (see [Testing](#testing)).

## Describe the change

Update private transactions page.
Add content around private nonce management.

## Issue fixed

fixes #849 

### For content changes

- [x] Doc content
- [ ] Doc pages organisation

### For tools changes

- [ ] CircleCI workflow
- [ ] Build and QA tools (lint, vale,…)
- [ ] MkDocs templates
- [ ] MkDocs configuration
- [ ] Python dependencies
- [ ] Node dependencies and JavaScript
- [ ] ReadTheDocs configuration
- [ ] GitHub integration

## Testing

https://hyperledger-besu--890.org.readthedocs.build/en/890/Concepts/Privacy/Private-Transactions/#private-nonce-management

### DIFF
--- a/docs/Concepts/Privacy/Private-Transactions.md
+++ b/docs/Concepts/Privacy/Private-Transactions.md
@@ -52,6 +52,7 @@ A nonce is the number of previous transactions made by the sender.
 the private transaction distributed to involved participants, and the privacy marker transaction (PMT) included on the
 public blockchain.
 Each of these transactions has its own nonce.
+The PMT nonce is the public nonce for the account.
 
 ### Private transaction nonce
 
@@ -64,11 +65,7 @@ That is, the nonce for account A for privacy group ABC is different to the nonce
     If sending more than one transaction for mining in the same block (that is, you aren't waiting for the transaction
     receipt), you must calculate the private transaction nonce outside Besu.
 
-### Privacy marker transaction nonce
-
-The PMT nonce is the public nonce for the account.
-
-### Nonce validation
+### Private nonce validation
 
 Unlike public transactions, private transactions are not submitted to the [transaction pool](../Transactions/Transaction-Pool.md).
 The private transaction is distributed directly to the participants in the transaction, and the PMT is submitted to the
@@ -99,3 +96,9 @@ The following private transaction flow illustrates when nonce validation occurs:
 
     If the private transaction was submitted with a correct nonce in step 1, the nonce is validated as correct.
     If an incorrect nonce was submitted, the private transaction execution fails.
+
+### Private nonce management
+
+You can call [`eth_getTransactionCount`](../../Reference/API-Methods.md#eth_gettransactioncount) to get the nonce, then
+use that nonce with [`eea_sendTransaction`](https://docs.ethsigner.consensys.net/en/stable/Reference/API-Methods/#eea_sendtransaction)
+to send a private transaction.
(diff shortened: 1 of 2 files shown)


macfarla (approved):
couple of suggestions

docs/Concepts/Privacy/Private-Transactions.md
```diff
@@ -10,99 +10,111 @@ description: Private transaction overview
...
+However, in this scenario, the private transaction execution fails when [processing the PMT](Private-Transaction-Processing.md)
+for the private transaction with the incorrect nonce.
 
-The nonce for the [privacy marker transaction] is the public nonce for the account.
+The following private transaction flow illustrates when nonce validation occurs:
 
-### Nonce validation
+1. Submit a private transaction with a [nonce value](#private-transaction-nonce).
+1. The private transaction is distributed to all participants in the privacy group.
+1. The PMT is created and submitted to the transaction pool with a nonce of `0` if using one time accounts.
```
rolandtyler, line 79:
```suggestion
1. The PMT is created and submitted to the transaction pool with a nonce of `0` if using one-time accounts.
```

rolandtyler (approved):
Minor suggestion, otherwise lgtm

PR #908, opened 2021-12-21
doc clique out-of-turn issue

### LINKED ISSUE #427: Clique "out of turn" potential issues
## Describe the bug

- [x] Missing content
- [ ] Outdated content
- [ ] Wrong content
- [ ] Confusing or misleading content
- [ ] Other

## Change suggestion

Include a section on "known issues , design constraints and potential defects" for various concensus models 

## additional content 
# Clique Issues

(Portions taken from a document from the clique repo or somewhere else on GitHub, GitHub links are my own , rest of content is from someone else)

Additional relevant details/repos

https://github.com/goerli/eips-poa/issues/6#issuecomment-485005551

https://github.com/goerli/eips-poa/issues/8

https://github.com/goerli/eips-poa/issues/7

https://github.com/goerli/eips-poa/issues/6[](url)

> source: https://docs.google.com/document/d/1pIW6Uac5Qanx_L5Y_G4Ucrx_gjITkBgzQKmlBQbW3Cg/edit

## Problem
The Görli network has on a number of occasions forked and become “stuck” on different forks both of which have the same total difficulty, so clients don’t re-org to the other side of the fork.

## Missing In-Turn Blocks
It’s believed that there are cases where an in-turn block is missed even though the validator is online.  Theoretically out of turn blocks should be published with a delay after the in-turn block should have been published, however the delay is quite small. Clock drift, malicious actors or buggy code may cause an out-of-turn validator to publish a new block sooner than expected. Alternatively the in-turn block could be slightly delayed due to clock drift on the in-turn validator or some performance problem causing a slight delay.

However it happens, if the in-turn validator receives the out-of-turn block before it publishes, it will import it into it’s chain and consider it the new canonical head (it’s the best block because the in-turn block doesn’t yet exist).  Typically getting a new chain head would cause the miner to stop mining it’s current block and start a new one based on the new chain head. This works well for PoW because the new block has higher difficulty and the best chance of being included in the canonical chain, however it’s the wrong behaviour for Clique because the new block will be out of turn whereas the original would have been in-turn and become the highest difficulty.

I’m not sure which clients exhibit this behaviour of cancelling the in-turn block creation but Pantheon definitely does and it looks like Geth does too (based on miner/miner.go line 86).

This behaviour can lead to the network forking. Imagine a network with four validators A, B, C and D.  The chain is at block 5 with total difficulty of 10 and block 6 should be produced by A. Ideally we should have block 6, produced by A with total difficulty of 12.

However, if B publishes an out-of turn block to A before it completes producing its block and A cancels creation of its block.  We then have:

A: height 6, TD 11
B: height 6, TD 11
C: height 5, TD 10
D: height 5, TD 10

C took longer to receive B’s block so after a delay publishes an out of turn block giving:
A: height 6, TD 11 (from B)
B: height 6, TD 11 (from B)
C: height 6, TD 11 (from C)
D: height 6, TD 11 (from D)

B should publish at height 7 but can’t because on its canonical chain it published block 6 and it doesn’t consider the other chain canonical so both chains will have to use an out of turn block for height 7 and thus have a TD of 12. With the validators evenly split the network is in a very bad state and will find it hard to recover. 

Q: Shouldn’t this resolve itself since eventually a block from the other fork should arrive first and given it has the higher TD it would then become canonical head? This fork should be a reasonably unstable state…

If A had published it’s in turn block, it would have had the highest total difficulty and the problem would be avoided.  

### Proposal: 
We should clarify the Clique spec to indicate that a validator should produce it’s in turn block even if it receives an out of turn block for that height (but not if the chain has progressed more than that one block?) There is work in progress now to apply this to Pantheon.

#### Out of Turn Delay Shorter Than Network Latency
The number of out of turn blocks in the network is significantly higher than necessary because the delay before publishing an out of turn block is randomly selected up to a fairly small maximum delay.  Whenever the randomly chosen delay is less than the block propagation time from the in-turn validator an unnecessary out-of-turn block will be published. At minimum this wastes resources. Potentially it could be contributing to the network forking.

If we have four validators, A, B, C and D.  If A & B are in Australia and C & D are in Europe then A & B will have low latency to each other but high latency to C & D. Thus, if out of turn blocks are produced by A and C at similar times and the same total difficulty, B will receive A’s block first and treat it as canonical and D will receive C’s block first. As they have the same total difficulty it doesn’t cause a re-org so the network remains forked. This may answer the question posed above about why random timing of received blocks doesn’t eventually cause the network to reach consensus.

#### Proposal:
Add two configurable options for the Clique algorithm:
OUT_OF_TURN_DELAY_MULTIPLIER
Currently explicitly 500ms in the spec
MIN_OUT_OF_TURN_DELAY
New concept so Rinkeby and Görli are effectively using 0ms
Absolute minimum amount of time to wait before publishing an out of turn block
Delay calculation is then: 
MIN_OUT_OF_TURN_DELAY + 
rand(SIGNER_COUNT * OUT_OF_TURN_DELAY_MULTIPLER)

On low latency networks the current settings may be enough but a geographically dispersed network like Görli may increase both options to ensure the out of turn delay exceeds the network latency and out of turn blocks aren’t created unnecessarily.

alexandratran:
## Pull request checklist

Use the following list to make sure your PR fits the Besu doc quality standard.

### Before creating the pull request

Make sure that:

- [x] [all commits in this PR are signed off for the DCO](https://wiki.hyperledger.org/display/BESU/DCO).
- [x] you read the [contribution guidelines](https://wiki.hyperledger.org/display/BESU/Contributing+to+documentation).
- [x] you have [tested your changes locally](https://wiki.hyperledger.org/display/BESU/MkDocs+And+Markdown+Guide#MkDocsAndMarkdownGuide-PreviewTheDocumentation) before submitting them to the community for review.

### After creating your pull request and tests finished

Make sure that:

- [x] you fixed all the issues raised by the tests, if any.
- [x] you verified the rendering of your changes on [ReadTheDocs.org PR preview](https://wiki.hyperledger.org/display/BESU/MkDocs+And+Markdown+Guide#MkDocsAndMarkdownGuide-PreviewwithReadTheDocs)
  and updated the testing link (see [Testing](#testing)).

## Describe the change

Doc limitation of using Clique.

## Issue fixed

fixes #427 

### For content changes

- [ ] Doc content
- [ ] Doc pages organisation

### For tools changes

- [ ] CircleCI workflow
- [ ] Build and QA tools (lint, vale,…)
- [ ] MkDocs templates
- [ ] MkDocs configuration
- [ ] Python dependencies
- [ ] Node dependencies and JavaScript
- [ ] ReadTheDocs configuration
- [ ] GitHub integration

## Testing

https://hyperledger-besu--908.org.readthedocs.build/en/908/HowTo/Configure/Consensus-Protocols/Clique/

### DIFF
--- a/docs/HowTo/Configure/Consensus-Protocols/Clique.md
+++ b/docs/HowTo/Configure/Consensus-Protocols/Clique.md
@@ -85,7 +85,7 @@ genesis file.
 
 In Clique, blocks created by in-turn validators are published immediately.
 Out-of-turn validators create blocks that are published after a short delay.
-Out-of-turn blocks have a lower difficulty than in-turn blocks, which allows small forks to resolve to the chain with
+In-turn blocks have a higher difficulty than out-of-turn blocks, which allows small forks to resolve to the chain with
 more in-turn blocks.
 
 However, when the out-of-turn delay is shorter than the block propagation delay, out-of-turn blocks may be published
--- a/docs/HowTo/Configure/Consensus-Protocols/Clique.md
+++ b/docs/HowTo/Configure/Consensus-Protocols/Clique.md
@@ -6,8 +6,8 @@ source: rinkeby.json
 
 # Clique
 
-Besu implements the Clique Proof-of-Authority (PoA) consensus protocol. The Rinkeby testnet uses
-Clique and private networks can also use Clique.
+Besu implements the [Clique](https://eips.ethereum.org/EIPS/eip-225) proof of authority (PoA) consensus protocol.
+The Rinkeby testnet uses Clique and private networks can also use Clique.
 
 In Clique networks, approved accounts, known as signers, validate transactions and blocks. Signers
 take turns to create the next block.
@@ -80,6 +80,19 @@ To connect to the Rinkeby testnet, start Besu with the
 node on a Clique private network, use the
 [`--genesis-file`](../../../Reference/CLI/CLI-Syntax.md#genesis-file) option to specify the custom
 genesis file.
+
+## Limitations
+
+In Clique, blocks created by in-turn validators are published immediately.
+Out-of-turn validators create blocks that are published after a short delay.
+Out-of-turn blocks have a lower difficulty than in-turn blocks, which allows small forks to resolve to the chain with
+more in-turn blocks.
+
+However, when the out-of-turn delay is shorter than the block propagation delay, out-of-turn blocks may be published
+before in-turn blocks.
+This may cause large, irresolvable forks in a network.
+
+We recommend using a more updated consensus protocol such as [IBFT 2.0](IBFT.md) or [QBFT](QBFT.md).
 
 <!-- Acronyms and Definitions -->
 
--- a/docs/HowTo/Configure/Consensus-Protocols/Clique.md
+++ b/docs/HowTo/Configure/Consensus-Protocols/Clique.md
@@ -92,7 +92,9 @@ However, when the out-of-turn delay is shorter than the block propagation delay,
 before in-turn blocks.
 This may cause large, irresolvable forks in a network.
 
-We recommend using a more updated consensus protocol such as [IBFT 2.0](IBFT.md) or [QBFT](QBFT.md).
+!!! important
+
+    We recommend using a more updated consensus protocol such as [IBFT 2.0](IBFT.md) or [QBFT](QBFT.md).
 
 <!-- Acronyms and Definitions -->
 
--- a/docs/HowTo/Configure/Consensus-Protocols/Clique.md
+++ b/docs/HowTo/Configure/Consensus-Protocols/Clique.md
@@ -7,7 +7,7 @@ source: rinkeby.json
 # Clique
 
 Besu implements the [Clique](https://eips.ethereum.org/EIPS/eip-225) proof of authority (PoA) consensus protocol.
-The Rinkeby testnet uses Clique and private networks can also use Clique.
+The Rinkeby and Goerli testnets uses Clique and private networks can also use Clique.
 
 In Clique networks, approved accounts, known as signers, validate transactions and blocks. Signers
 take turns to create the next block.


docs/HowTo/Configure/Consensus-Protocols/Clique.md
```diff
@@ -6,8 +6,8 @@ source: rinkeby.json
 
 # Clique
 
-Besu implements the Clique Proof-of-Authority (PoA) consensus protocol. The Rinkeby testnet uses
-Clique and private networks can also use Clique.
+Besu implements the [Clique](https://eips.ethereum.org/EIPS/eip-225) proof of authority (PoA) consensus protocol.
+The Rinkeby testnet uses Clique and private networks can also use Clique.
```
garyschulte, line 10: It might be worth mentioning Goerli uses clique also

garyschulte (approved):
LGTM, non-blocking suggestion about Goerli

rolandtyler (approved):
lgtm

docs/HowTo/Configure/Consensus-Protocols/Clique.md
```diff
@@ -81,6 +81,19 @@ node on a Clique private network, use the
...
+In Clique, blocks created by in-turn validators are published immediately.
+Out-of-turn validators create blocks that are published after a short delay.
+In-turn blocks have a higher difficulty than out-of-turn blocks, which allows small forks to resolve to the chain with
+more in-turn blocks.
+
+However, when the out-of-turn delay is shorter than the block propagation delay, out-of-turn blocks may be published
+before in-turn blocks.
+This may cause large, irresolvable forks in a network.
+
+We recommend using a more updated consensus protocol such as [IBFT 2.0](IBFT.md) or [QBFT](QBFT.md).
```
bgravenorst, line 95: Probably worth highlighting this using "important" block?

PR #1208, opened 2022-11-30
Updating public network info for sync 

non-fungible-nelson:
Added time and size estimates. Removed warning on snap sync restarts (besu now resumes from last point). Moved fast sync to the bottom and added not on deprecation on Mainnet. 

Signed-off-by: Matt Nelson <85905982+non-fungible-nelson@users.noreply.github.com>

### DIFF
--- a/docs/public-networks/get-started/connect/sync-node.md
+++ b/docs/public-networks/get-started/connect/sync-node.md
@@ -46,7 +46,7 @@ All times are hardware dependent. We tabulate these running AWS instances m6gd.2
 | Fast        | ~1.5 days   | Average Disk |
 | Full        | ~weeks      | Most Disk Usage | 
    
-As of late 2022, an average Mainnet Snap Sync will consume around ~650GB on Bonsai. While the worldstate is syncing, Besu will download and import the blockchain in the background. It will need to catch up to the current chain head and sync the worldstate to participate on Mainnet. 
+As of late 2022, an average Mainnet Snap Sync will consume around ~750GB on Bonsai. While the worldstate is syncing, Besu will download and import the blockchain in the background. It will need to catch up to the current chain head and sync the worldstate to participate on Mainnet. 
 
 ### Snap synchronization
 
--- a/docs/public-networks/get-started/connect/sync-node.md
+++ b/docs/public-networks/get-started/connect/sync-node.md
@@ -76,11 +76,11 @@ You can't switch from fast sync to snap sync.
 If your node is blocked in the middle of a fast sync, you can start over using snap sync instead by stopping the node,
 deleting the data directory, and starting over using `--sync-mode=X_SNAP`.
 
-See [how to read the Besu metrics charts](../../how-to/monitor/understand-metrics.md) when using snap sync.
-
 You can restart Besu during a snap sync in case of hardware or software problems.
 The sync resumes from the last valid world state and continues to download blocks starting from the
 last downloaded block.
+
+See [how to read the Besu metrics charts](../../how-to/monitor/understand-metrics.md) when using snap sync.
 
 ### Checkpoint synchronization
 
@@ -110,10 +110,6 @@ number, and total difficulty as in the following example.
       "totalDifficulty": "0xA2539264C62BF98CFC6"
     }
     ```
-You can restart Besu during a checkpoint sync in case of hardware or software problems.
-The sync resumes from the last valid world state and continues to download blocks starting from the
-last downloaded block.
-
 
 !!! note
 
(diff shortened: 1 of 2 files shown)


alexandratran (approved):
Made some additional edits and rearranged some content. LGTM now!

docs/public-networks/get-started/connect/sync-node.md
```diff
@@ -4,39 +4,132 @@ description: Full and archive node types
...
+## Sync times
+
+To sync with a public network, Besu runs two processes in parallel: the world state sync and the
+blockchain download.
+
+The following table shows the average world state sync time for each sync mode.
+All times are hardware dependent; this table is based on running AWS instances m6gd.2xlarge.
+Each sync mode also has its own world state database size.
+
+| Sync mode   | Time to sync world state | Disk usage    |
```
siladu, line 30: Can be a second PR, but I think it would be even better to put total initial sync times here since this is what the users will care most about.
Worldstate would be good as well, but maybe as a drill-down.

I am certain that some users don't understand (and even don't care to understand) the details of the sync, they are just looking for an overall time estimate so they can plan and know if it's not going well.

I worry that users will see a number on a nice table without reading the text and assume checkpoint sync only takes 5 hours :)

docs/public-networks/get-started/connect/sync-node.md
```diff
@@ -4,39 +4,132 @@ description: Full and archive node types
...
+| ----------- | ------------------------ | ------------- |
+| Snap        | ~6 hours                 | Average disk  |
+| Checkpoint  | ~5 hours                 | Smallest disk |
+| Fast        | ~1.5 days                | Average disk  |
+| Full        | ~weeks                   | Largest disk  |
```
siladu, line 35: Suggest we use the actual options here for clarity/consistency, e.g. X_CHECKPOINT

siladu: Can be put disk size estimates in the third column?

docs/public-networks/get-started/connect/sync-node.md
```diff
@@ -4,39 +4,132 @@ description: Full and archive node types
...
+Archive nodes require more disk space than full nodes.
 
-For Besu on Mainnet, archive nodes [require more disk space](../../concepts/data-storage-formats.md#storage-requirements)
-than full nodes.
+## Sync times
+
+To sync with a public network, Besu runs two processes in parallel: the world state sync and the
+blockchain download.
+
+The following table shows the average world state sync time for each sync mode.
```
siladu, line 26: Call out that this table is for mainnet?

Maybe could have another table for goerli/sepolia? I think most stakers test on goerli so they can try out setting up a validator. 

Or maybe could just say "testnets will take significantly less time and space to sync"?

docs/public-networks/get-started/connect/sync-node.md
siladu: I'm sure we could get numbers for these based on recent sync tests

docs/public-networks/get-started/connect/sync-node.md
```diff
@@ -4,39 +4,132 @@ description: Full and archive node types
...
+!!! note
+
+    If using [Clique](../../../private-networks/how-to/configure/consensus/clique.md) consensus, the
+    checkpoint must be the beginning of an epoch.
+
+If you enable checkpoint sync without a checkpoint configuration in the genesis file, Besu snap
+syncs from the genesis block.
+
+You can restart Besu during a checkpoint sync in case of hardware or software problems.
+The sync resumes from the last valid world state and continues to download blocks starting from the
```
siladu, line 123: woohoo!

docs/public-networks/get-started/connect/sync-node.md
```diff
@@ -4,39 +4,132 @@ description: Full and archive node types
...
+last downloaded block.
 
 ### Fast synchronization
 
+!!! important
+
+    It might become impossible to sync Ethereum Mainnet using fast sync in the future.
+    Update Besu to a version that supports newer sync methods.
```
siladu, line 131: It's a good callout, but I don't think FAST sync will happen again once the node is already in sync...so I think you only really need to update the config if you ever need to resync

siladu (commented):
A few suggestions, but could be handled in separate PRs. Won't block this PR on it as what you have is good and certainly an improvement.

alexandratran:
Addressed some of @siladu's feedback in this PR and also raised https://github.com/hyperledger/besu-docs/issues/1211 to add more numbers in the future.

PR #1224, opened 2022-12-20
Update fast-sync-min-peers description

### LINKED ISSUE #1192: Document pivot block selector on PoS networks
From https://github.com/hyperledger/besu/pull/4488
On PoS network we use as pivot block, the last finalized block sent by the Consensus Layer, so we do
not need peers, and so all the logic for selecting the pivot block from peers
has been moved from `FastSyncActions` to `PivotSelectorFromPeers`.
We do not need anymore the `TransictionPeerSelector`, and the `--fast-sync-min-peers`
applies only to PoW networks.

alexandratran:
Update `fast-sync-min-peers` description to note that it only applies to PoW networks. Also fix the the `genesis-file` option description right after it.

Fixes #1192.

### DIFF
--- a/docs/public-networks/reference/cli/options.md
+++ b/docs/public-networks/reference/cli/options.md
@@ -724,22 +724,15 @@ Contact email address to send to the Ethstats server specified by [`--ethstats`]
     fast-sync-min-peers=8
     ```
 
-The minimum number of peers required before starting [fast synchronization](../../get-started/connect/sync-node.md#fast-synchronization).
+The minimum number of peers required before starting [fast synchronization](../../get-started/connect/sync-node.md#fast-synchronization)
+in [proof of work](../../how-to/use-pow/mining.md) networks.
 The default is 1.
 
-!!! note
+!!! important
 
-    If synchronizing in `FAST` mode, most historical world state data is unavailable. Any methods
-    attempting to access unavailable world state data return `null`.
+    This option only applies to proof of work networks.
 
 ### `genesis-file`
-
-Use the genesis file to create a custom network.
-
-!!!tip
-
-    To use a public Ethereum network such as Goerli, use the [`--network`](#network) option. The
-    network option defines the genesis file for public networks.
 
 === "Syntax"
 
@@ -765,11 +758,11 @@ Use the genesis file to create a custom network.
     genesis-file="/home/me/me_node/customGenesisFile.json"
     ```
 
-The path to the genesis file.
+The path to the [genesis file](../../concepts/genesis-file.md).
 
 !!!important
 
-    You cannot use the [`--genesis-file`](#genesis-file) and [`--network`](#network) options at the
+    You can't use the [`--genesis-file`](#genesis-file) and [`--network`](#network) options at the
     same time.
 
 ### `graphql-http-cors-origins`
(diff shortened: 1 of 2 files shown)


docs/public-networks/reference/cli/options.md
```diff
@@ -724,23 +724,16 @@ Contact email address to send to the Ethstats server specified by [`--ethstats`]
     fast-sync-min-peers=8
     ```
 
-The minimum number of peers required before starting [fast synchronization](../../get-started/connect/sync-node.md#fast-synchronization).
+The minimum number of peers required before starting [fast synchronization](../../get-started/connect/sync-node.md#fast-synchronization)
+in [proof of work](../../how-to/use-pow/mining.md) networks.
 The default is 1.
```
fab-10, line 729: The default is 5

docs/public-networks/reference/cli/options.md
```diff
@@ -724,23 +724,16 @@ Contact email address to send to the Ethstats server specified by [`--ethstats`]
...
-!!! note
+!!! important
 
-    If synchronizing in `FAST` mode, most historical world state data is unavailable. Any methods
-    attempting to access unavailable world state data return `null`.
+    This option only applies to proof of work networks.
 
 ### `genesis-file`
 
-Use the genesis file to create a custom network.
```
fab-10, line 737: the changes to `genesis-file` section are unrelated, have they been inserted by error?

docs/public-networks/reference/cli/options.md
```diff
@@ -724,23 +724,16 @@ Contact email address to send to the Ethstats server specified by [`--ethstats`]
...
-The minimum number of peers required before starting [fast synchronization](../../get-started/connect/sync-node.md#fast-synchronization).
+The minimum number of peers required before starting [fast synchronization](../../get-started/connect/sync-node.md#fast-synchronization)
+in [proof of work](../../how-to/use-pow/mining.md) networks.
 The default is 1.
 
-!!! note
+!!! important
 
-    If synchronizing in `FAST` mode, most historical world state data is unavailable. Any methods
-    attempting to access unavailable world state data return `null`.
```
fab-10, line 733: This part could be kept, and instead of says `FAST` mode, it could be changes to something like "any mode different from `FULL`"

fab-10 (changes requested):

docs/public-networks/reference/cli/options.md
alexandratran: No, I noticed that the description in of `genesis-file` was redundant and mis-formatted, so I fixed it in this PR as well (as per the PR description).

docs/public-networks/reference/cli/options.md
alexandratran: Does this note make sense to put here in the description of `fast-sync-min-peers` though? I'm unsure on how it's related. Would it be better under the [`sync-mode` option](https://besu.hyperledger.org/en/stable/public-networks/reference/cli/options/#sync-mode), or in the [fast sync section](https://besu.hyperledger.org/en/stable/public-networks/get-started/connect/sync-node/#fast-synchronization)?

I notice that the same note is also in [this block parameter section](https://besu.hyperledger.org/en/stable/public-networks/how-to/use-besu-api/json-rpc/?h=block+paramet#block-parameter). I'm not sure it fits well here either.

fab-10: I agree it make more sense to move it under the sync mode option

fab-10 (approved):

rolandtyler (approved):
LGTM

PR #1279, opened 2023-03-06
Add link to quorum plugins

alexandratran:
Add link to Quorum Plugins doc site from the `--security-module` CLI option.

See https://github.com/ConsenSys/doc.quorumplugins/issues/118


bgravenorst (approved):
LGTM

PR #1349, opened 2023-06-23
Update versioning

alexandratran:
Update versioning to make the [release process](https://wiki.hyperledger.org/display/BESU/Documentation+release+process) simpler.

- Rename the `stable` version to the version number, so creating new versions doesn't require renaming the previous version.
- Keep the `stable` label and path for user-friendliness and so footer links and external links to the docs don't need to be changed.
- Remove duplicate versioned docs.


joshuafernandes (approved):
LGTM

PR #1365, opened 2023-07-20
Add note to explain that more than one `--from` argument can be passed to the `import` sub command

matthew1001:


docs/public-networks/reference/cli/subcommands.md
```diff
@@ -57,6 +57,8 @@ You can specify the starting index of the block range to import with `--start-bl
 
 You can specify the ending index (exclusive) of the block range to import with `--end-block`. If omitted, all blocks after the start block will be imported.
 
+You can specify multiple `--from` arguments. This can be useful when blocks have been exported over time to multiple files.
```
shemnon, line 60: Maybe clarify that the order of the block files will be the CLI order?

shemnon (approved):

docs/public-networks/reference/cli/subcommands.md
matthew1001: Sure, that's a good point. Let me add that under a new commit.

matthew1001: Added commit e1df688aa93603b83b8c224fb354b8d6e17338cc @shemnon

docs/public-networks/reference/cli/subcommands.md
```diff
@@ -57,6 +57,8 @@ You can specify the starting index of the block range to import with `--start-bl
 
 You can specify the ending index (exclusive) of the block range to import with `--end-block`. If omitted, all blocks after the start block will be imported.
 
+You can specify multiple `--from` arguments. This can be useful when blocks have been exported over time to multiple files. If multiple files are provided they will be read in the order specified in the command.
```
alexandratran, line 60: To specify multiple `--from` arguments, do you separate the arguments with commas or do you specify the `--from` option multiple times?

matthew1001: @alexandratran you specify it multiple times. I did wonder about adding more than 1 `--from` argument into the example in the docs, if you think that would be useful?

alexandratran: Yeah, that would be helpful!

matthew1001: Done :) See new commit 2bf1add6c18c47ae820d6fed0b94eec8945697b9

alexandratran (approved):

docs/public-networks/reference/cli/subcommands.md
alexandratran, line 60:
```suggestion
You can specify the ending index (exclusive) of the block range to import with `--end-block`. If omitted, all blocks after the start block are imported.

You can specify multiple `--from` arguments. This can be useful when blocks have been exported over time to multiple files. If multiple files are provided they are read in the order specified in the command.
```

PR #1367, opened 2023-07-28
Update examples on GraphQL page

### LINKED ISSUE #1352: Clarify the graphql results to expect when trying the example
In the graphql docs at https://besu.hyperledger.org/stable/public-networks/how-to/use-besu-api/graphql the first example shows the retrieval of the sync status for the node:

image

In certain cases (I think when the node is fully synced?) the query in the example returns `syncing: null` in the response. I think it would be useful to add a note to the docs, perhaps just after the "GraphQL requests with GraphiQL app" screenshot, indicating that the query may return

```
{
  "data" : {
    "syncing" : null
  }
}
```

in some cases so users aren't concerned that their Besu config isn't incorrect.

I wonder if it would be useful to add a more basic example before the sync-state example, such as:

`--data '{ "query": "{block{number}}"}'` which returns e.g.

```
{
  "data" : {
    "block" : {
      "number" : "0x281"
    }
  }
}
```

or `--data '{ "query": "{gasPrice}"}'` which returns e.g.

```
{
  "data" : {
    "gasPrice" : "0x0"
  }
}
```

alexandratran:
Added simpler examples on GraphQL page and added note about syncing request.

fixes #1352 

preview: https://besu-docs-bc3echm75-hyperledger.vercel.app/development/public-networks/how-to/use-besu-api/graphql


macfarla (approved):
LGTM

matthew1001:
Thanks @alexandratran, the preview page looks great to me.

PR #1372, opened 2023-08-03
Updated topics to be both public & private

### LINKED ISSUE #1371: Add public networks tag to Use API pages
All the subpages under the **Public networks > How to > [Use the Besu API](https://besu.hyperledger.org/stable/public-networks/how-to/use-besu-api)** section should have both the **public networks** and **private networks** tags at the top of the page, but they only have the **private networks** tag.

Add the **public networks** tag to each of these pages. See how to use [Docusaurus doc tags](https://docusaurus.io/docs/create-doc#doc-tags).

mjsmike62:
Update the Public networks > How to > Use the Besu API topics with both the public networks tag so that now these topics have both the public networks and private networks tags.

Fixes #1371


alexandratran (approved):
Looks good to me.

PR #1398, opened 2023-09-07
ignore env/ dir

macfarla:
I think this is created when doing npm commands for releases


bgravenorst (approved):
Ah, thanks. This folder can sometimes be annoying.

PR #1454, opened 2023-11-22
Clarified behaviour of eth_call strict parameter

### LINKED ISSUE #1429: eth_call strict clarification
Re https://github.com/hyperledger/besu/issues/5943

Confusion arising from the fact that we [said](https://github.com/hyperledger/besu/blob/main/CHANGELOG.md#2110-breaking-changes)

> eth_call will not fail for account balance issues by default.

which implies the default for strict is false - whereas the code actually tries to get a "sensible" default based on the other value - so check the balance if gas != 0 (even if strict not specified)

the docs currently say

> By default, eth_call does not fail if the sender account has an insufficient balance. This is done by setting the balance of the account to a large amount of ether. To enforce balance rules, set the [strict parameter](https://besu.hyperledger.org/public-networks/reference/api/objects#transaction-call-object) in the transaction call object to true.

But actually, if you specify gas, the account balance will be checked, even if you do not specify `strict`

Suggestion -

If you specify `strict`: true, the balance will be checked - call will fail if the sender account has an insufficient balance to send the transaction with specified gas parameters
If you specify `strict`: false, the balance will not be checked, regardless of gas params
If you do not specify `strict`, the sender account balance will be checked against gas parameters if supplied.
If you do not want the sender's account balance to be checked, send zero gas or specify strict: false

code reference for specifics https://github.com/hyperledger/besu/blob/f58f6cffca1c877b566ee5f6e980a53b352bd3dc/ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/internal/methods/EthCall.java#L170

MadelineMurray:
Fixes #1429


macfarla (approved):
excellent

alexandratran (approved):

PR #1476, opened 2024-01-06
Fix Typo "debug_resyncWorldstate" accordingly

ladidan:
**Issue**

Trying to initiate a world state resync given the current docs triggers a "message":"Method not found""

**Fix**

capitalise "S" in "WorldState" according to the given RPC methods in

ethereum/api/src/main/java/org/hyperledger/besu/ethereum/api/jsonrpc/RpcMethod.java


bgravenorst (approved):
Approved

PR #1478, opened 2024-01-09
sync-min-peers is not only for fast sync

macfarla:
refs besu PR adding alias sync-min-peers for fast-sync-min-peers https://github.com/hyperledger/besu/pull/6372


bgravenorst (approved):
lgtm

PR #1479, opened 2024-01-09
Updated environment variable example for bonsai-historical-block-limit

macfarla:
environment variable still had the old name

also added a missing space on landing page


bgravenorst (approved):
lgtm

PR #1515, opened 2024-02-18
SNAP and Bonsai are now Besu defaults

macfarla:
A few changes to Besu defaults coming in 24.2.0 -

SNAP and Checkpoint no longer "early access" - so removed references to X_SNAP and X_CHECKPOINT
https://github.com/hyperledger/besu/pull/6405
Snap sync is now default for named networks
https://github.com/hyperledger/besu/pull/6530
Bonsai is now default storage format
https://github.com/hyperledger/besu/pull/6536

### DIFF
--- a/docs/public-networks/get-started/connect/mainnet.md
+++ b/docs/public-networks/get-started/connect/mainnet.md
@@ -58,7 +58,7 @@ Run the following command or specify the options in a [configuration file](../..
 
 ```bash
 besu \
-  --sync-mode=X_SNAP           \
+  --sync-mode=SNAP           \
   --data-storage-format=BONSAI \
   --rpc-http-enabled=true      \
   --rpc-http-host=0.0.0.0      \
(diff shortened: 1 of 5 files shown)


docs/public-networks/get-started/connect/mainnet.md
```diff
@@ -58,7 +58,7 @@ Run the following command or specify the options in a [configuration file](../..
 
 ```bash
 besu \
-  --sync-mode=X_SNAP           \
+  --sync-mode=SNAP           \
```
bgravenorst, line 61: Hey @macfarla, should we still specify `--sync-mode` and  `--data-storage-format` in the command (since it's now the default).

bgravenorst (approved):
Just one question, otherwise lgtm.

docs/public-networks/get-started/connect/mainnet.md
macfarla: yeah good point - removed

cryptohazard:
hello,

the change of the default storage format is not updated on this doc: https://besu.hyperledger.org/public-networks/concepts/data-storage-formats#forest-of-tries-vs-bonsai-tries

I somehow missed that change in the CHANGELOG when upgrading to 24.3.0 from 23.X

PR #1539, opened 2024-03-13
adding java install and upgrade page

non-fungible-nelson:


alexandratran (approved):
Made some edits, LGTM.

PR #1544, opened 2024-03-21
Update block parameter `pending` and `blockHash`

### LINKED ISSUE #1520: Document "pending" for qbft_getValidatorsByBlockNumber
Document support for "pending" block string. See https://github.com/hyperledger/besu/pull/6436 for more information and context:

> I think https://besu.hyperledger.org/public-networks/how-to/use-besu-api/json-rpc#block-parameter needs updating to say "Use only with [eth_getTransactionCount](https://besu.hyperledger.org/public-networks/reference/api#eth_gettransactioncount) and [qbft_getValidatorsByBlockNumber](https://besu.hyperledger.org/private-networks/reference/api#qbft_getvalidatorsbyblocknumber)"

alexandratran:
This PR adds `qbft_getValidatorsByBlockNumber` as a method that supports the `pending` value for block parameter, and adds a description of what `pending` means for `getValidators`. (Fixes #1520)

Also adds note on any methods that include `pending` to specify that `pending` returns the same value as `latest` (except for `eth_getTransactionCount` and `qbft_getValidatorsByBlockNumber`).

Also removes `blockHash` from the block parameter description of methods except those specified as supporting `blockHash` on the [block parameter page](https://besu.hyperledger.org/public-networks/how-to/use-besu-api/json-rpc#block-parameter).

### DIFF
--- a/docs/private-networks/reference/api/index.md
+++ b/docs/private-networks/reference/api/index.md
@@ -1135,7 +1135,13 @@ For private contracts, `priv_call` is the same as [`eth_call`](../../../public-n
 
 - `call`: _object_ - [transaction call object](../../../public-networks/reference/api/objects.md#transaction-call-object)
 
-- `blockNumber`: _string_ - hexadecimal or decimal integer representing a block number, or one of the string tags `latest`, `earliest`, `finalized`, or `safe`, as described in [block parameter](../../../public-networks/how-to/use-besu-api/json-rpc.md#block-parameter)
+- `blockNumber`: _string_ - hexadecimal or decimal integer representing a block number, or one of the
+  string tags `latest`, `earliest`, `pending`, `finalized`, or `safe`, as described in
+  [block parameter](../../../public-networks/how-to/use-besu-api/json-rpc.md#block-parameter)
+
+  :::note
+  `pending` returns the same value as `latest`.
+  :::
 
 #### Returns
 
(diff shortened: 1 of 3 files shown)


matthew1001:
Thanks for these updates @alexandratran. The removal of `pending` from the various `clique_`/`ibft_`/`qbft_` methods does make sense in that it doesn't have any meaning for them (except `qbft_getValidatorsByBlockNumber` obviously), but it is actually supported on all of those methods, it's just that it returns the same value as if you'd passed in `latest`. Whether or not that's the right thing for the code to do I'm not sure, but currently it's the case so I'd suggest we make the docs reflect the current code behaviour. Apologies, I probably wasn't very clear on that when I mentioned the suggested docs changes in the original PR.

So I'm wondering whether it's best to leave `pending` on the other `clique_`/`ibft_`/`qbft_` but with a caveat that `pending` returns the same value as `latest`. What do you think?

I think the update to https://besu.hyperledger.org/public-networks/how-to/use-besu-api/json-rpc#block-parameter to say "use only with eth_getTransactionCount and qbft_getValidatorsByBlockNumber" is correct as it only has meaning for those methods. I'm wondering if we need to update what the `pending` string means though as it has a different meaning for `eth_getTransactionCount` as it does for `qbft_getValidatorsByBlockNumber`. For the former it's "The last block mined plus pending transactions." but for the latter it's "The list of validators that will be used to produce the next block.". Given that this is in the `public-networks` section in the docs I'm not sure how much detail we want to go into on that page. I wonder if the docs for `qbft_getValidatorsByBlockNumber` (https://github.com/hyperledger/besu-docs/blob/5fbd1781e0d5e6db697a4a7d355048b1c1a03d4c/docs/private-networks/reference/api/index.md?plain=1#L2320) should describe what `pending` does and the public docs page should say something like "use only with eth_getTransactionCount and qbft_getValidatorsByBlockNumber. See qbft)getValidatorsByBlockNumber` for detail about its use."?

Sorry it's a bit complicated. I'm not sure what's going to be clearest to the user given the split between public and private docs and the fact that all the other methods accept `pending` but don't actual do anything different to `latest`.

alexandratran:
@matthew1001 Thanks for the clarifications. I added `pending` back to the `clique_`/`ibft_`/`qbft_` methods, with a note that `pending` returns the same value as `latest` (except `qbft_getValidatorsByBlockNumber`. Also updated the block parameter section and the `qbft_getValidatorsByBlockNumber` description to include the `getValidators` meaning of `pending`.

matthew1001:
Thanks @alexandratran that looks much clearer.

It looks like you've removed pending from a number of the other methods (e.g. `eth_getBlockTransactionCountByNumber`). These all support "pending" the same as the other methods (i.e. Besu accepts it but returns the same as for "latest") so I think we probably still want `pending` listed in the options for the other methods.

Other than that it looks good. I'll stick an approval on if you're happy to address that last comment.

matthew1001 (approved):
Approved, I think we just need to re-instate `pending` on the other methods as pretty much all of them accept it, they just return the same as `latest`

bgravenorst (approved):
lgtm

PR #1555, opened 2024-04-02
Add receipt compaction option

### LINKED ISSUE #1533: Document `--receipt-compaction-enabled`
Document the feature flag `--receipt-compaction-enabled`. It allows users to disable receipt compaction if needed, for instance in situations where performance of RPCs is really important and the compacted format might introduce overhead.

**Context:** A new compact receipt encoding has been introduced to reduce the storage requirements for receipts in a synced mainnet snap sync node. This reduction is achieved by omitting the bloom filter from receipts and trimming leading zeros from log topics and log data. Versioning  for the database format has been added for backwards compatibility. Once a user upgrades to this version, they cannot downgrade to a previous version without losing data. Besu will support both the existing and the new compacted receipt formats. The new format is enabled by default, optimizing storage without requiring manual intervention.  A new database metadata format, version 3, is introduced for both the forest and bonsai database models.

**See the following for more context:**
- https://github.com/hyperledger/besu/pull/6602

joaniefromtheblock:
Closes #1533


docs/public-networks/reference/cli/options.md
```diff
@@ -2943,6 +2943,46 @@ random-peer-priority-enabled=true
...
+
+```bash
+receipt-compaction-enabled=true
+```
+
+</TabItem>
+
+</Tabs>
+
+Enables or disables receipt compaction. Compacting receipts reduces storage by trimming unnecessary data from transaction receipts. Enabled by default, this option optimizes storage efficiency. It can be disabled if the compaction process introduces performance overhead.
```
siladu, line 2984: > Enabled by default

Currently it's disabled by default. Also, not released yet, but might be soon.

siladu: > It can be disabled if the compaction process introduces performance overhead.

I'm not sure if this this performance overhead was actually observed but would defer to @jframe. Thinking we don't need to suggest why you might want to disable it, maybe could remove this sentence?

jframe: Agree we should remove the sentence about performance. At the moment it's not clear without additional testing in high load environments.

joaniefromtheblock: Hi @jframe and @siladu . I updated this PR. Is it ready for release?

jframe (approved):

alexandratran (approved):
One suggestion for consistency otherwise lgtm.

PR #1579, opened 2024-04-17
Update redirects

### LINKED ISSUE #1527: Add redirect rule for versioned doc root paths
Linking to the root path of a versioned doc yields a "Page Not Found" error (ex: https://besu.hyperledger.org/23.10.3/).

This issue is to add a redirect rule to [`vercel.json`](https://github.com/hyperledger/besu-docs/blob/main/vercel.json) to redirect versioned doc root paths to the introduction pages for each versioned doc. Will require using regex matching (see [Vercel redirect docs](https://vercel.com/docs/projects/project-configuration#redirects)).

alexandratran:
This PR fixes some redirect issues, adding redirects to the `@docusaurus/plugin-client-redirects` and removing `vercel.json` because this site is not hosted on Vercel. Partially fixes #1527 (versioned doc root redirects may not be possible due to the client-side redirects not working with regex).


bgravenorst (approved):
lgtm

PR #1612, opened 2024-06-10
add dco workflow

macfarla:
this acts as an alternative to DCO bot, and fills the same requirement. You can have both.


joaniefromtheblock:
Thanks @macfarla

PR #1655, opened 2024-07-17
Allow version update with workflow dispatch

cdivitotawela:
## Description
Was looking to trigger the version update workflow by re-creating the release 24.7.0. However when a release it re-created from the same commit which has been created a release, it does not seems to trigger the even type released. Added workflow_dispatch also to the same workflow so that users can manually trigger the version update if necessary.


joaniefromtheblock (approved):
LGTM

PR #1694, opened 2024-09-23
Remove version update workflow

cdivitotawela:
## Description
Besu docs no longer maintain versions in the code. Hence removed the workflow that perform version update changes. Also updated the main workflow by removing the dependency of links job as there is no links job anymore


joshuafernandes (approved):
LGTM

joshuafernandes (approved):
LGTM

PR #1716, opened 2024-10-18
use only vale spelling via the GHA

joshuafernandes:
## Description
- Use only vale spelling via the GHA
- remove cspell


m4sterbunny (approved):
LGTM

PR #1730, opened 2024-11-21
Privacy deprecation

macfarla:
## Description

Add deprecation warnings for privacy pages. 

### Issue(s) fixed
Related to besu changes https://github.com/hyperledger/besu/pull/7897

Fixes #

### Preview

-

PR #1734, opened 2024-11-22
added deprecation notice to CLI options for privacy

macfarla:
## Description
follow-on from #1730 to deprecate specific CLI options


alexandratran (approved):
Fixed links and other minor edits. LGTM.

PR #1739, opened 2024-11-26
Remove --Xsnapsync-bft-enabled

pullurib:
### Issue(s) fixed
[#7924](https://github.com/hyperledger/besu/issues/7924)

Fixes #


docs/private-networks/reference/cli/options.md
```diff
@@ -922,6 +922,48 @@ privacy-url="http://127.0.0.1:8888"
 
 The URL on which the [Tessera node](../../tutorials/privacy/index.md#3-create-tessera-configuration-files) is running.
 
+### `snapsync-bft-enabled`
```
macfarla, line 925: do we want to add the version of besu that this becomes available eg (since 24.12.0)

macfarla:
ok per comments on the besu PR this option is being removed altogether

docs/public-networks/reference/cli/options.md
```diff
@@ -5589,15 +5589,6 @@ The synchronization mode. Use `SNAP` for [snap sync](../../concepts/node-sync.md
...
 
-:::warning Early access feature
-
-`--Xsnapsync-bft-enabled` is an early access feature available in Besu version 24.7.1 and later.
-It is not stable, so use this option with caution.
-
-Use `--Xsnapsync-bft-enabled` with `--sync-mode=SNAP` to enable snap sync in QBFT and IBFT 2.0 private networks.
-
-:::
-
```
macfarla, line 5600: so if the option is being removed altogether, we want this change (deletion)

docs/private-networks/reference/cli/options.md
```diff
@@ -922,6 +922,48 @@ privacy-url="http://127.0.0.1:8888"
...
+```
+
+</TabItem>
+
+</Tabs>
+
+Enables or disables snapsync in IBFT 2.0 and QBFT private networks. The default is `false`.
+
+Use `--snapsync-bft-enabled` with `--sync-mode=SNAP` to enable snapsync in QBFT and IBFT 2.0 private networks.
+
```
macfarla, line 966: so if the option is being removed altogether, we DON'T want this change (addition)

macfarla (changes requested):

docs/public-networks/reference/cli/options.md
pullurib: Based on the feedback on other PR, the option is being marked as deprecated now and will be removed in future release. Will update the docs accordingly.

docs/private-networks/reference/cli/options.md
pullurib: I agree, will cleanup

macfarla (commented):
I think we could just remove this content altogether when this option is no longer needed (deprecated)

alexandratran (approved):
I removed all mentions of `--Xsnapsync-bft-enabled` because snap sync is now supported without the experimental flag.

macfarla (approved):
LGTM but might be good to wait until https://github.com/hyperledger/besu/pull/7930 is merged - right now you do still need to use `--Xsnap-bft-enabled` if you want to use SNAP on BFT networks (then again, anyone who needs to use that flag prob already knows about it)

macfarla:
still waiting on the Besu PR - apologies for jumping the gun on this one. I'll make this one draft for now.

PR #1741, opened 2024-11-27
removed PKI and TLS for P2P

macfarla:
## Description
PKI was removed a while back
TLS for P2P experimental only; being removed now

relates to besu PR https://github.com/hyperledger/besu/pull/7942

Fixes #

### Preview

-


alexandratran (commented):
[removed]

alexandratran (approved):
LGTM.

PR #1754, opened 2024-12-18
deprecate k8s nat

joshuafernandes:
## Description
- deprecated K8S NAT
- new tutorial for cross K8S cluster across cloud providers


bgravenorst (approved):
lgtm

PR #1791, opened 2025-03-31
Document optional `debug_setHead` parameter.

### LINKED ISSUE #1784: Document optional parameter for `debug_setHead`
Document new optional parameter for `debug_setHead` that allows you to move the head worldstate along with the blockchain. See https://github.com/hyperledger/besu/pull/7821 for more information.

bgravenorst:
## Description
Document optional `debug_setHead` parameter.

- 

### Issue(s) fixed

Fixes #1784 

### Preview
https://besu-docs-git-fork-bgravenorst-doc-1784-hyperledger.vercel.app/public-networks/reference/api#debug_sethead


docs/public-networks/reference/api/index.md
```diff
@@ -1435,17 +1435,30 @@ curl -X POST --data '{"jsonrpc":"2.0","method":"debug_resyncWorldState","params"
 
 ### `debug_setHead`
 
-Sets the current head of the local chain to the block matching the specified block number.
+Sets the local chain head to the specified block. Optionally, moves the [bonsai](../../concepts/data-storage-formats.md#bonsai-tries)
+world state to that block.
+
+Moving the world state allows expensive operations like [`debug_traceBlock`](#debug_traceblock)
+to run on historical blocks without replaying all intermediate states. This is helpful to avoid
+out of memory errors when executing RPC calls on historical states.
```
garyschulte, line 1443: This is good, but perhaps we can be more clear that it is specific to the option for moving the worldstate only.  The caution call out is good also.

garyschulte (approved):
Approving with expectation that we can explicitly call out that the motivation for moving the worldstate only applies if that option is enabled

docs/public-networks/reference/api/index.md
```diff
@@ -1435,17 +1435,30 @@ curl -X POST --data '{"jsonrpc":"2.0","method":"debug_resyncWorldState","params"
...
+
+:::caution
+This method should not be used when a consensus client is directing Besu, or while the node is
+actively importing or proposing blocks.
```
garyschulte, line 1447: Maybe this can be a bit scarier also.  It is likely to cause database corruption if the node is trying to import or process blocks at all.  Not just execution engine, any change to the chain head or worldstate from any consensus mechanism will cause problems while this call is executing.

One other callout here might be that the RPC operation might time out if the distance the head is moving is more than 5000 blocks or so.  The operation will continue to run on besu, but the RPC response itself might time out

alexandratran (approved):
Minor suggestions.

PR #1806, opened 2025-04-30
use absolute paths in examples

macfarla:
## Description
~ gets resolved relative to current directory which is confusing - so it's not a good example to use.

### Issue(s) fixed

Fixes #

### Preview

-


bgravenorst (approved):
lgtm

Issue #1722, opened 2024-10-22
Make option syntax clear and consistent


alexandratran:
There are some inconsistencies with how we present the syntax of various CLI options in the reference. For example:

> `--rpc-http-tls-cipher-suite=<cipherSuiteName>[, <cipherSuiteName>...]`

> `--bootnodes[=<enode://id@host:port>[,<enode://id@host:port>...]...]`

> `--net-restrict=<subnet>[,<subnet>,...]`

> `--banned-node-ids=<bannedNodeId>[,<bannedNodeId>...]...`

These are some inconsistent examples of how we denote comma-separated lists, and there are still other variants. Also, we currently use square brackets to denote an optional value, but some options might require using square brackets for arrays, which could cause confusion. We should decide on a consistent and clear way to document these types of options.

m4sterbunny:
Note on the proposed t/f solution: no go!
there are instances where the [= is correct. 
For example "--bonsai-limit-trie-logs-enabled"
LLM looking at the besu code tells me: fallbackValue = "true" means it can be used as a flag (without a value) and when used as a flag, it defaults to true
(If arity = "1" were set, it would require a value)
Tests also show it used as a flag:
So given that --bonsai-limit-trie-logs-enabled can be used as a flag; that means docs should give [=<true|false>]

PR #1880, opened 2025-11-28
1722 consistent syntax

m4sterbunny:
## Description
- Standardizes the syntax for True / False configs
- Adds missing defaults

### Issue(s) fixed

Partial fix for #1722 

### Preview

-


macfarla (approved):
happy for this change. 

I did notice at least one spot where we have a third way - but this could be addressed in a follow up PR 

`--era1-import-prepipeline-enabled[=<Boolean>]`

PR #1890, opened 2025-12-04
standardizes syntax for t/f and makes clear when flag applies

m4sterbunny:
## Description
Clean up / verification of T/F
Boolean - 2 variants remain
required =<true|false>
optional [=<true|false>], where default is false and passing the boolean is optional, the bash example shows a flag only

### Issue(s) fixed

Fixes #partial 1772

### Preview

-


m4sterbunny:
Closing this: there is no single logic route to identify from code whether flag is accepted where arity is not 1. Only testing can confirm. Will reduce the scope and adjust the original issue.

PR #1891, opened 2025-12-11
cleans up t/f syntax and applies requred for arity=1 booleans

m4sterbunny:
## Description

Cleans up t/f syntax to either <true|false> (when value must be passed with option) or [=<true|false>] (when flag is sufficient)

> Warning: this does NOT act to confirm all settings are correct per ,. vs. []. It DOES confirm that where arity=1 which is a reliable method to know that value is required, then is documented as <> required.

The list of those options that are documented and those that may be valid but not yet documented is:

```
auto-log-bloom-caching-enabled (arity is 1 and set to required <true|false>)
bonsai-limit-trie-logs-enabled
color-enabled
discovery-enabled (arity is 1 and set to required <true|false>)
engine-jwt-disabled
engine-rpc-enabled
era1-import-prepipeline-enabled
genesis-state-hash-cache-enabled
graphql-http-enabled
graphql-mtls-enabled
history-expiry-prune
json-pretty-print-enabled
metrics-enabled
metrics-push-enabled
p2p-enabled (arity is 1 and set to required <true|false>)
plugin-continue-on-error (arity is 1 and set to required <true|false>)
random-peer-priority-enabled
receipt-compaction-enabled (arity not 1 but is set as required <true|false>, not edited)
remote-connections-limit-enabled
revert-reason-enabled
rpc-http-authentication-enabled
rpc-http-enabled
rpc-http-tls-ca-clients-enabled
rpc-http-tls-client-auth-enabled
rpc-http-tls-enabled
rpc-ws-authentication-enabled
rpc-ws-enabled
rpc-ws-ssl-client-auth-enabled
rpc-ws-ssl-enabled
snapsync-server-enabled
snapsync-synchronizer-pre-checkpoint-headers-only-enabled
snapsync-synchronizer-transaction-indexing-enabled
strict-tx-replay-protection-enabled
tx-pool-enable-balance-check
tx-pool-enable-save-restore
tx-pool-no-local-priority
version-compatibility-protection

Not documented, arity = 1--> confirm is live and is boolean:

peerDiscoveryEnabled
poaDiscoveryRetryBootnodes
compress
dnsEnabled
dnsUpdateEnabled
natMethodFallbackEnabled
enableOptimizedOpcodes
```

### Issue(s) fixed

Fixes #Partial for 1722

### Preview

-

### DIFF
--- a/docs/public-networks/reference/cli/options.md
+++ b/docs/public-networks/reference/cli/options.md
@@ -218,7 +218,9 @@ auto-log-bloom-caching-enabled=false
 
 </Tabs>
 
-Enables or disables automatic log bloom caching. APIs such as [`eth_getLogs`](../api/index.md#eth_getlogs) and [`eth_getFilterLogs`](../api/index.md#eth_getfilterlogs) use the cache for improved performance. The default is `true`.
+Enables or disables automatic log bloom caching. APIs such as [`eth_getLogs`](../api/index.md#eth_getlogs) and [`eth_getFilterLogs`](../api/index.md#eth_getfilterlogs) use the cache for improved performance. 
+
+The default is `true`.
 
 If automatic log bloom caching is enabled and a log bloom query reaches the end of the cache, Besu
 performs an uncached query for logs not yet written to the cache.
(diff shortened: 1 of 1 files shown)


bgravenorst (approved):
lgtm

PR #1912, opened 2026-02-18
1910 reload plugins

### LINKED ISSUE #1910: Update method to allow reloading configuration for all plugins with a single call
Extend [`plugins_reloadplugin`](https://besu.hyperledger.org/public-networks/reference/api#plugins_reloadpluginconfig) method, to make the plugin name parameter optional, and in case it is not passed then all enabled plugins are reloaded.

See: https://github.com/hyperledger/besu/pull/9563

m4sterbunny:
## Description

- plugins reload param is optional

### Issue(s) fixed

Fixes #1910 

### Preview

-


fab-10 (approved):

macfarla (commented):
suggestions on wording

PR #1959, opened 2026-04-29
Improve navigation between public and private networks

### LINKED ISSUE #1955: Improve navigation between public and private networks
- Investigate methods of displaying public/private overlapping content in both sidebars, instead of listing the links in the index pages as [we do currently](https://besu.hyperledger.org/private-networks/how-to).
- Ensure top nav label is highlighted when you're in each section.

alexandratran:
## Description

- Update styles for top nav links to make it more clear which section you're in
- Remove tags (the links are pretty useless) in favor of breadcrumbs
- Swizzle breadcrumbs to include "Public networks" vs "Private networks" in the hierarchy (by default these are omitted because they use different sidebars)

Note: In the future, we may consider adding `_partials` files for overlapping content, to avoid these awkward index pages on the private network side: https://besu.hyperledger.org/private-networks/how-to. For now, let's table that large change until we have a better idea of the direction of our private network offering.

### Issue(s) fixed

Fixes #1955 

### Preview

- https://besu-docs-6bdin23zc-hyperledger.vercel.app/public-networks

### DIFF
--- a/docs/private-networks/concepts/index.md
+++ b/docs/private-networks/concepts/index.md
@@ -2,8 +2,6 @@
 title: Concepts
 description: private networks concepts overview
 sidebar_position: 3
-tags:
-  - private networks
 ---
 
 # Concepts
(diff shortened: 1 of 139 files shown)


bgravenorst (approved):
lgtm

PR #2005, opened 2026-05-18
Add info for Besu HSM plugin

alexandratran:
## Description

Add information for HSM-backed node/validator keys, referencing the Besu HSM plugin.

### Preview

https://besu-docs-4yrg136ug-hyperledger.vercel.app/private-networks/how-to/configure/consensus/qbft

### DIFF
--- a/docs/private-networks/how-to/configure/consensus/qbft.md
+++ b/docs/private-networks/how-to/configure/consensus/qbft.md
@@ -27,8 +27,7 @@ It might take significant time to recover after nodes are restarted.
 
 :::tip HSM-backed validator keys
 
-For QBFT validators, the validator address is derived from the
-[node public key](../../../../public-networks/concepts/node-keys.md#node-public-key).
+For QBFT validators, [node addresses](../../../../public-networks/concepts/node-keys.md) are validator addresses.
 To store a validator's node key in a Hardware Security Module (HSM) instead of
 on disk, use a security module plugin, such as the
 [Besu HSM plugin](https://github.com/besu-eth/besu-hsm-plugin), with the
--- a/docs/private-networks/how-to/configure/consensus/qbft.md
+++ b/docs/private-networks/how-to/configure/consensus/qbft.md
@@ -25,9 +25,15 @@ It might take significant time to recover after nodes are restarted.
 
 :::
 
-:::tip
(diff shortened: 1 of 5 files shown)


docs/public-networks/reference/cli/options.md
```diff
@@ -6446,30 +6446,32 @@ Specify the required [`--rpc-ws-ssl-truststore-file`](#rpc-ws-ssl-truststore-fil
...
 ```
 
 </TabItem>
 
 </Tabs>
 
-Name of the security module plugin to use. For example, a Hardware Security Module (HSM) or V3 filestore plugin.
+Name of the security module plugin to use for [node key](../../concepts/node-keys.md) storage.
+For example, use a Hardware Security Module (HSM) or V3 filestore plugin, such as the
+[Besu HSM plugin](https://github.com/besu-eth/besu-hsm-plugin).
```
usmansaleem, line 6474: nit: Either here on on line 6476, we can mention that the default value of `--security-module` is `localfile` and the location of private key file can be override using `--node-private-key-file`.

usmansaleem (approved):
LGTM - minor comment.

docs/private-networks/how-to/configure/consensus/qbft.md
```diff
@@ -25,9 +25,15 @@ It might take significant time to recover after nodes are restarted.
...
 :::
 
-:::tip
-
-You can use a plugin to securely store a validator's key using the [`--security-module`](../../../../public-networks/reference/cli/options.md#security-module) option.
+:::tip HSM-backed validator keys
+
+For QBFT validators, the validator address is derived from the
+[node public key](../../../../public-networks/concepts/node-keys.md#node-public-key).
+To store a validator's node key in a Hardware Security Module (HSM) instead of
```
bgravenorst, line 31: Do we need to be specific and mention this is to avoid storing the private node key on disk?

bgravenorst (approved):
Approved, but I do have a question regarding specifically calling out the private key for clarity.

docs/private-networks/how-to/configure/consensus/qbft.md
alexandratran: Updated wording to hopefully avoid confusion of public vs private key.

PR #2018, opened 2026-05-28
Clarify totalDifficulty pre-merge behavior

homika482005:
## Description

Updated the `totalDifficulty` field description in the Besu API objects documentation to clarify that it is only returned for pre-merge blocks.

## Related issue(s)

Relates to  besu-eth/besu#9836


bgravenorst (approved):
lgtm 🚀 
Thanks for the contribution.

PR #100, opened 2019-11-24
Updated priv_getTransactionReceipt to add new fields

MadelineMurray:
Updated priv_getTransactionReceipt to add new fields

PR #2038, opened 2026-06-17
Add IBFT2->QBFT migration guide

### LINKED ISSUE #2009: Document how to migrate from IBFT to QBFT
Add section or guide on how to migrate from IBFT to QBFT, using the `startBlock` config item.

alexandratran:
## Description

Add guide for migrating a live network from IBFT 2.0 to QBFT consensus.

## Related issue(s)

Fixes #2009 

## Preview

https://besu-docs-git-fork-alexandratran-2009-ibft-migrate-hyperledger.vercel.app/private-networks/how-to/configure/consensus/migrate-from-ibft-to-qbft


jframe (commented):
Documentation looks good. But when I tested this I found some bugs that made this feature unusable. So I think we should hold off adding docs until these are fixed.

The IBFT2 and QBFT RPCs are broken https://github.com/besu-eth/besu/issues/10679
And IBFT2 blocks are still built after the merge https://github.com/besu-eth/besu/issues/10680

docs/private-networks/how-to/configure/consensus/migrate-from-ibft-to-qbft.md
```diff
@@ -0,0 +1,115 @@
...
+If any node is still using the old genesis configuration at the migration block, that node will fork
+from the network.
+:::
+
+### 4. Verify the migration
+
+When the network reaches the migration block, Besu handles the cutover from IBFT 2.0 to QBFT 
+automatically.
+Confirm that the network is producing QBFT blocks by calling
+[`qbft_getValidatorsByBlockNumber`](../../../reference/api.md#qbft_getvalidatorsbyblocknumber):
```
jframe: This RPC method can't be used to verify that QBFT blocks are being produced. It will work even when IBFT blocks are still being produced. Maybe a bug?

I think only way to know it has switched to QBFT is to check the logs. At the specified block number for QBFT to start there will be a log line like this
```QbftBesuControllerBuilder | Imported block #100```. This actual log line will specify if it's a empty block or not and the number txs, gas etc.

jframe (approved):

PR #2068, opened 2026-07-09
Update config options and subcommands reference

alexandratran:
## Description

This PR makes several formatting changes to make the Besu reference more intuitive. In both public and private networks:

- Move options and subcommands out of the "Besu command line" category and into the root reference folder, making it faster to navigate to each page. Links throughout the docs are updated to reflect this restructure, + redirects.
- Retitle the options page to "Configuration options" since they are often referred to colloquially as "configs," and they're often used in config files, not only on the command line. Simplify the intro text + mention that these options are part of the command line interface (CLI), in case users search for "command line" or "CLI." Add same clarification to subcommands reference.
- In the options page, remove existing h2 headings so the option headings can be promoted to h2. This simplifies and makes the TOC more readable.
- Remove the "Syntax" tab, so the remaining three tabs reflect examples for each way the option can be specified (command line, env var, config file).
- Add a horizontal line between each config option and each top-level subcommand, to slightly improve readability.

## Preview

https://besu-docs-git-fork-alexandratran-cli-hyperledger.vercel.app/public-networks/reference/options


bgravenorst (approved):
lgtm

